PR #11, opened 2022-05-20
10 Update workflow for automatic upload bundle

### LINKED ISSUE #10: Enhance workflow for automatically update models
After a PR is merged, we need to manually check which bundles are updated/added, and download the large files if necessary, and then compress each bundle into a zip file and upload to github release. After finishing all these steps, the `model_info.json` file should also be updated.
It would save a lot of time if all these steps can be finished automatically.

yiheng-wang-nv:
Fixes #10 .

### Description
This PR is used to add the automatic uploading actions after a PR is merged. The workflow is like:
flow chart

### Status
**Ready**


yiheng-wang-nv:
Hi @wyli @Nic-Ma and @ericspod , I just marked the status of this PR into "Ready", and could you please help to review it?
I attached a flow chart at the beginning, and it might be helpful to understand the processes.

During a previous commit, I used the spleen ct segmentation bundle for test: [log](https://github.com/Project-MONAI/model-zoo/runs/6876008372?check_suite_focus=true), [uploaded place](https://github.com/Project-MONAI/model-zoo/releases/tag/hosting_storage_v1), [produced new branch](https://github.com/Project-MONAI/model-zoo/tree/11-auto-update-model-info
)

wyli:
thanks, this looks great, does this support deletion as well? just wanted to confirm that if I submit a PR to permanently remove a bundle folder, the PR won't break the pipeline. And I guess we should provide some instructions to the users/admins about how to remove a bundle folder.

Nic-Ma:
Hi @wyli ,

That's a good corner case, @yiheng-wang-nv please make sure a PR can delete some bundles in the source code(not in the release storage).

Thanks.

.github/update_model_info.py
```diff
@@ -10,117 +10,124 @@
...
-    actual_hash_func = look_up_option(hash_type.lower(), SUPPORTED_HASH_TYPES)
-    return actual_hash_func()
+    # step 1
+    bundle_path = os.path.join(models_path, bundle_name)
+    if not os.path.exists(bundle_path):
+        return (False, "Bundle path not exist")
 
+    # temp_dir = "test_tmp"
+    # if not os.path.exists(temp_dir):
+    #     os.makedirs(temp_dir)
```
Nic-Ma: Please remove the comment test code.

Thanks.

Nic-Ma (commented):
Hi @wyli ,

Could you please help review the PR? I think you are expert on the Github workflows.

Thanks in advance.

.github/update_model_info.py
```diff
@@ -10,117 +10,124 @@
...
 
-def update_model_info(models_path: str, model_info_file: str = "model_info.json"):
+    # step 2
+    temp_dir = tempfile.mkdtemp()
+    temp_path = os.path.join(temp_dir, bundle_name)
```
ericspod, line 46: You could put every line in this function after this one within one large `try/finally` block with `shutil.rmtree(temp_dir)` as the only line in the finally block. This would ensure that the directory is cleaned up no matter where the exception happens and means you only need the rmtree line once.

ericspod:
The version strings are part of the compressed file's names so we should mention in the specification that versions should follow semantic versioning and contain only characters valid in filenames.

yiheng-wang-nv:
> thanks, this looks great, does this support deletion as well? just wanted to confirm that if I submit a PR to permanently remove a bundle folder, the PR won't break the pipeline. And I guess we should provide some instructions to the users/admins about how to remove a bundle folder.

Thanks! @wyli and @Nic-Ma . If delete a whole bundle, according to the logic (`utils.get_changed_bundle_list`, it is used to grab all changed and exsiting bundles), the github release part will not be changed actually. I just pushed a new commit to print some information to address this case.

.github/update_model_info.py
```diff
@@ -10,118 +10,124 @@
...
+    if len(bundle_list) > 0:
+        for bundle in bundle_list:
+
+            # create a temporary copy of the bundle for further processing
+            temp_dir = tempfile.mkdtemp()
+            update_state, msg = update_model_info(
+                bundle_name=bundle, temp_dir=temp_dir, models_path=models_path, model_info_file=model_info_file
+            )
+            shutil.rmtree(temp_dir)
+
```
yiheng-wang-nv, line 116: Thanks @ericspod . I modify the code structure of the main function, and put the remp dir logic here. What do you think?

README.md
```diff
@@ -65,4 +65,9 @@ Ideally, the new branch should be based on the latest `dev` branch.
 1. Reviewer and contributor may have discussions back and forth until all comments addressed.
 1. Wait for the pull request to be merged.
 
+### Remove a bundle
+
+If an existing bundle needs to be removed, please follow the steps mentioned above to do the delete manipulations during the created new branch in your fork, and then create a new pull request.
+However, bundles that are already stored in the release will not be changed if the corresponding source code is removed.
+
```
yiheng-wang-nv, line 72: Hi @wyli and @Nic-Ma , as for the instructions, I wrote like this in the readme, what do you think?

.github/update_model_info.py
ericspod: I meant for the `rmtree` part to stay in the function so that you ensure you have cleanup, otherwise an exception before this line will cause the cleanup to be missed. The body of the method would be something like this:

```python
def do_something():
...
    try:
        # create temp dir
        ... # do work
        return whatever
    finally:
        shutil.rmtree(temp_dir)
```

README.md
wyli: thanks, looks good to me, perhaps add: "Please create an issue or contact the repository maintainers if a released bundle should be removed permanently."

requirements.txt
```diff
@@ -1 +1,2 @@
-monai>=0.8.1
+monai>=0.9.0
+gdown>=4.4.0
```
wyli:
```suggestion
monai[gdown]>=0.9.0
```

wyli (approved):
looks good to me, a minor point --  requirements.txt could be updated

.github/update_model_info.py
ericspod: The way you have the code now will work and we'll not likely have an issue with it, the `try/finally` way ensures cleanup no matter if you have an exception or not so it's more of a guarantee. I'm fine with how it is now though.

Nic-Ma:
> The version strings are part of the compressed file's names so we should mention in the specification that versions should follow semantic versioning and contain only characters valid in filenames.

Hi @ericspod ,

I created a small PR for it: https://github.com/Project-MONAI/MONAI/pull/4512, could you please help review it?
And I think this PR is ready to merge now.

Thanks in advance.

PR #22, opened 2022-05-27
init blossom ci

YanxuanLiu:
Init Blossom CICD workflow


pxLi (approved):
LGTM. Please trigger some dummy PR builds after merged, thx~

Nic-Ma:
Hi @wyli @yiheng-wang-nv ,

Could you please help review this PR? I think you have much experience on the Github workflow and Blossom.

Thanks in advance.

pxLi:
> Hi @wyli @yiheng-wang-nv ,
> 
> Could you please help review this PR? I think you have much experience on the Github workflow and Blossom.
> 
> Thanks in advance.

this is the template from blossom, we can only modify the allowed users part. (blossom will do integrity check for all other parts, it will fail the check if we modified anything else). Just make sure the github IDs are correct, and we should be good~

yiheng-wang-nv (approved):

yiheng-wang-nv (approved):

PR #31, opened 2022-06-08
Mednist GAN Bundle

ericspod:
Fixes # .

### Description
This is the bundle for the MedNIST GAN network based off the example notebooks in the tutorials repo. I've followed the code pretty closely with that found in the notebooks with a few changes that don't impact results. There are two scripts for training and inference. 

I have as the license the copyright information that is at the top of our files and notebooks but I'm open to changing this to something more appropriate. 

I've also included the weights for the network as a pt file directly rather than as a download link because it's so small. I feel we're ok with tiny networks to do this but it can be changed if that's better.

### Status
**Ready**


yiheng-wang-nv:
> I wasn't sure what the license should be for this one so yes some help on it would be good.

Hi @ericspod , Nic and I discussed about the license, and could you please refer to the first cell of this notebook: https://colab.research.google.com/drive/1wy8XUSnNWlhDNazFdvGBHLfdkGvOHBKe and use the same license?

models/mednist_gan/configs/inference.json
```diff
@@ -0,0 +1,89 @@
...
+        }
+    ],
+    "evaluator": {
+        "_target_": "SupervisedEvaluator",
+        "device": "@device",
+        "val_data_loader": "@dataloader",
+        "network": "@network",
+        "inferer": "@inferer",
+        "postprocessing": "@postprocessing",
+        "prepare_batch": "$lambda batchdata, *_,**__: (batchdata.to(@device),None,(),{})",
```
Nic-Ma, line 82: Hi @ericspod ,

I think maybe our default `prepare_batch` can support the GAN case?
https://github.com/Project-MONAI/MONAI/blob/dev/monai/engines/utils.py#L131
And the tutorial is here:

models/mednist_gan/configs/inference.json
```diff
@@ -0,0 +1,89 @@
...
+        "val_handlers": "@handlers"
+    },
+    "inferring": [
```
Nic-Ma, line 85: Should we use the word "inference" here? I am not very sure.

Thanks.

models/mednist_gan/configs/metadata.json
```diff
@@ -0,0 +1,59 @@
...
+    "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
+    "version": "0.1.0",
+    "changelog": {
+        "0.1.0": "Initial version"
+    },
+    "monai_version": "0.9.0",
+    "pytorch_version": "1.10.0",
+    "numpy_version": "1.21.0",
+    "optional_packages_version": {
+        "pytorch-ignite": "0.4.8"
```
Nic-Ma: I think this bundle also uses `PIL` for image reading?

Thanks.

models/mednist_gan/configs/metadata.json
```diff
@@ -0,0 +1,59 @@
...
+    "data_source": "MedNIST dataset kindly made available by Dr. Bradley J. Erickson M.D., Ph.D. (Department of Radiology, Mayo Clinic)",
+    "data_type": "jpeg",
+    "network_data_format": {
+        "inputs": {
+            "latent": {
+                "type": "tuples",
+                "format": "latent",
+                "num_channels": 0,
+                "spatial_shape": [
+                    64
```
Nic-Ma, line 28: [64, 64]?

Thanks.

models/mednist_gan/configs/train.json
```diff
@@ -0,0 +1,168 @@
+{
+    "imports": [
+        "$from functools import partial",
+        "$import glob",
+        "$import numpy as np",
+        "$import torch",
```
Nic-Ma: No need to import numpy and torch as they are dependent packages of MONAI.

Thanks.

models/mednist_gan/configs/train.json
```diff
@@ -0,0 +1,168 @@
...
+        "d_loss_function": "$partial(discriminator_loss, disc_net=@dnetwork)",
+        "d_prepare_batch": "$prepare_batch",
+        "d_train_steps": 5,
+        "g_update_latents": true,
+        "latent_shape": "@latent_size",
+        "key_train_metric": "$None",
+        "train_handlers": "@handlers"
+    },
+    "training": [
+        "$@gnetwork.conv.add_module('activation', torch.nn.Sigmoid())",
```
Nic-Ma, line 161: I think you added this activation layer here because the loss function doesn't have activation arg?
Maybe it's better to still use the `postprocessing` arg?

Thanks.

models/mednist_gan/scripts/losses.py
```diff
@@ -0,0 +1,26 @@
+import torch
+
+disc_loss_criterion = torch.nn.BCELoss()
+gen_loss_criterion = torch.nn.BCELoss()
+real_label = 1
+fake_label = 0
+
+
+def prepare_batch(batchdata, device=None, non_blocking=False):
+    return batchdata["image"].to(device=device, non_blocking=non_blocking)
```
Nic-Ma: We already have this function in: 
https://github.com/Project-MONAI/MONAI/blob/dev/monai/engines/utils.py#L130

Thanks.

ericspod:
> Hi @ericspod , Nic and I discussed about the license, and could you please refer to the first cell of this notebook: https://colab.research.google.com/drive/1wy8XUSnNWlhDNazFdvGBHLfdkGvOHBKe and use the same license?

The content in that cell isn't a license, it's stating that the data is licensed under CC but just points to that license's text. The license that should be here is for the network itself, not the data. I've stated the data's licensing conditions in the README.md file so that should suffice for that, but we don't want to assert that the network is copyrighted by the same author as the data and distributed under their terms. What I put initially in the license file is our copyright header but that isn't sufficient either, so the next best thing I would say is to use the Apache 2.0 license MONAI uses?

models/mednist_gan/configs/inference.json
ericspod: We've used the active tense in other places ("training") so this follows the pattern.

models/mednist_gan/configs/metadata.json
ericspod: No this is the generator network which accepts a latent vector of 64 values as input.

models/mednist_gan/configs/train.json
ericspod: Sorted.

models/mednist_gan/configs/inference.json
ericspod: It can be made to work by changing the inference dataset to be a list of dictionaries but the we'd have to use the `CommonKeys.IMAGE` key which isn't accurate. The dataset right now is a list of latent inputs: `"$[torch.rand(@latent_size) for i in range(@num_samples)]"`. I would have to change this to `"$[{'image': torch.rand(@latent_size)} for i in range(@num_samples)]"` to make `default_prepare_batch` work which expects dictionaries, but calling the latent vectors "image" is a lie really. I had thought to use this also as an example of how to do a custom function like this.

models/mednist_gan/configs/metadata.json
ericspod: Added.

models/mednist_gan/configs/train.json
ericspod: I had this here because it's what's in the original [notebook in the tutorials repo](https://github.com/Project-MONAI/tutorials/blob/main/modules/mednist_GAN_workflow_dict.ipynb). I thought that sticking to it in this way would be better so those familiar with the notebook could follow here where the same components go, and also as an example of modifying the network. The inference notebook uses the postprocessing activation, I can change that here still if you want.

models/mednist_gan/scripts/losses.py
ericspod: Yeah that works with all the keys changed to "reals".

models/mednist_gan/configs/inference.json
Nic-Ma: OK, I see your point.

Thanks.

models/mednist_gan/configs/metadata.json
Nic-Ma: Oh, I see.

Thanks.

models/mednist_gan/configs/train.json
Nic-Ma: OK, I see your point.

Thanks.

Nic-Ma:
Hi @wyli ,

Could you please help share some comments about the license question?
I am also not sure what licenses we should put for this bundle.
Others look good to me now.

Thanks in advance.

Nic-Ma:
Hi @ericspod ,

About the license, I think maybe you can refer to the BERT model in the huggingface:
https://huggingface.co/bert-base-uncased/blob/main/LICENSE
We may need some guide or policy for developers to write license?

Thanks.

ericspod:
I think we do need to consider what our licensing should be and what to recommend to contributors. The BERT link uses Apache 2.0 like MONAI, I'd be fine with that. We could also use MIT license that's more permissive and I'd be fine with that as well, but we should definitely establish a policy.

Nic-Ma:
Hi @ericspod ,

We discussed the license topic with @pdogra89 and the team last week, and updated the contribution guide for it:
https://github.com/Project-MONAI/model-zoo/blob/dev/CONTRIBUTING.md#license-format

Could you please help update this PR for this last part?

Thanks in advance.

ericspod:
Yes I'll get on updating this, I had wanted to revisit what to do with licensing.

ericspod:
I've added a statement into the license regarding the MedNIST data, this goes with the license for the network itself. Does this meet the discussed license criteria and the contribution guidelines?

Nic-Ma (approved):
Should we put the `model.pt` in the large_files to reduce the repo size even it's less than 25MB?
Others look good to me.

Thanks for the update.

models/mednist_gan/configs/metadata.json
```diff
@@ -0,0 +1,60 @@
+{
+    "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
```
yiheng-wang-nv: Hi @ericspod , thanks for the updates. As for the `schema` here, it may still need to be modified, since it requires key `image` within `inputs`, but `latent` is used here.
There is not error raised here, since the `verify_metadata` printed the log rather than raised error here. The related place is just modified (see: https://github.com/Project-MONAI/MONAI/pull/4686).
Therefore, should we prepare a new `schema` here?

models/mednist_gan/configs/metadata.json
```diff
@@ -0,0 +1,60 @@
...
+    "task": "Generate random hand images from the MedNIST dataset",
+    "description": "This example of a GAN generator produces hand xray images like those in the MedNIST dataset",
+    "authors": "MONAI Team",
+    "copyright": "Copyright (c) MONAI Consortium",
+    "intended_use": "This is an example of a GAN with generator discriminator networks using MONAI, suitable for demonstration purposes only.",
+    "data_source": "MedNIST dataset kindly made available by Dr. Bradley J. Erickson M.D., Ph.D. (Department of Radiology, Mayo Clinic)",
+    "data_type": "jpeg",
+    "network_data_format": {
+        "inputs": {
+            "latent": {
```
yiheng-wang-nv, line 23: According to the latest changes of premerge CI tests, here could you please help to add this bundle into: https://github.com/Project-MONAI/model-zoo/blob/11952118b61ac64404f841153dbefec531baa90c/ci/bundle_custom_data.py#L16 ? This bundle may not need to be tested with input shape. Thanks!

ericspod:
> Should we put the `model.pt` in the large_files to reduce the repo size even it's less than 25MB? Others look good to me.
> 
> Thanks for the update.

The file is so tiny I don't think it would make much of a difference, I want to have at least one example of a whole bundle in the repo without having to explain our large file system.

models/mednist_gan/configs/metadata.json
ericspod: The issue is that the input isn't an image but a latent vector of values so it wouldn't be correct to call it that. We're going to have other networks eventually that don't take images as input but instead have one or more inputs like a latent vector or 1D signals. What can we do to modify the schema to allow for arbitrarily-named inputs and outputs that's backwards compatible with the current schema?

models/mednist_gan/configs/metadata.json
ericspod: OK that's done.

models/mednist_gan/configs/metadata.json
yiheng-wang-nv: Hi @ericspod , I added a new schema (https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_flexible_inputs_20220718.json) which removes the "required" validation keyword in "inputs" (other places are the same as the current schema). For this schema, users do not have to provide "images". What do you think?
cc @Nic-Ma

ericspod: This will work for validation but then we aren't really checking anything now. Perhaps we're limited by what is possible in the schema but we would want to be able to say that the "properties" section contains one or more items of with arbitrary non-duplicate names having the structure given currently in "image". For now we can use this new schema to get this bundle finalised if anyone else is good with that.

yiheng-wang-nv: Hi @ericspod , according to our last meeting's discussion, I uploaded your schema in: https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_generator_20220718.json
could you please use this schema so far for this bundle? Then this PR can be merged. I will create a ticket about the new schema design, and we can enhance that part in the future.

yiheng-wang-nv: https://github.com/Project-MONAI/model-zoo/issues/65

ericspod: That's done, we do need to have a think about how a general schema will work.

PR #40, opened 2022-06-24
39 Add premerge ci tests

### LINKED ISSUE #39: Adding premerge ci tests
We can enable the following ci tests for a pull request if it changes (adds) a bundle:

1. If `large_files` file exists, check the download link and checksum for each large file.
2. Verify the metadata format.
3. Verify the data shape of network.
4. Check TorchScript export, if the bundle supports.

yiheng-wang-nv:
Fixes #39 .

### Description
This PR is used to add some premerge ci tests for verifying changed bundles. All tests will be executed on blossom.
In addition, this PR also adds `einops` into the `swin_unetr_btcv_segmentation` bundle since this lib is required.

### Status
**Ready**


.github/workflows/verify_bundle.yml
```diff
@@ -0,0 +1,18 @@
+name: verify-bundle
+
```
yiheng-wang-nv: This file is used for demo, later we will move the action into blossom. FYI @YanxuanLiu @Nic-Ma

ci/verify_bundle.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def verify_metadata_format(bundle_path: str):
+    """
+    This function is used to verify the metadata format.
+
+    """
+    verify_cmd = "python -m monai.bundle verify_metadata --meta_file configs/metadata.json --filepath eval/schema.json"
+
+    subprocess.check_call(f"{verify_cmd}", shell=True, cwd=bundle_path)
```
wyli: If there are some testing errors, subprocess.check_call might not print detailed error messages, this could increase the difficulty of debugging. So perhaps we change to use native python commands for this?

wyli (approved):
Thanks, it looks good to me.

PR #67, opened 2022-07-29
66-update brats bundle

KumoLiu:
Signed-off-by: KumoLiu <yunl@nvidia.com>

Fixes #66  .

### Description
update brats bundle for 0.9.1

### Status
**Ready**

### Please ensure all the checkboxes:

- [x] Codeformat tests passed locally by running `./runtests.sh --codeformat`.
- [ ] In-line docstrings updated.
- [x] Update `version` and `changelog` in `metadata.json`.
- [x] Ensure versions of packages such as `monai`, `pytorch` and `numpy` are correct in `metadata.json`.
- [ ] Descriptions should be consistent with the content, such as `eval_metrics` of the provided weights and TorchScript modules.
- [ ] Files larger than 25MB are excluded and replaced by providing download links in `large_file.yml`.

PR #74, opened 2022-08-03
Revert "66-update brats bundle"

yiheng-wang-nv:
Reverts Project-MONAI/model-zoo#67

The PR has some unsolved issues, thus we need to revert it so far.


Nic-Ma (approved):

yiheng-wang-nv:
Update: @KumoLiu will submit a new PR to fix the torch version issue, thus this PR is closed.

PR #83, opened 2022-08-08
add detection bundle

Can-Zhao:

Fixes # .

### Description
Add lung nodule detection bundle
Readme and meta.json not finished

### Status

- [ ] Codeformat tests passed locally by running `./runtests.sh --codeformat`.


yiheng-wang-nv:
Hi @Can-Zhao , the schema check is failed (https://github.com/Project-MONAI/model-zoo/runs/7832105295?check_suite_focus=true#step:2:219), according to the schema this bundle used, the following properties are required:
```
"type",
"format",
"num_channels",
"spatial_shape",
"dtype",
"value_range"
```

models/lung_nodule_ct_detection/configs/train.json
```diff
@@ -0,0 +1,450 @@
...
+            ],
+            [
+                10,
+                10,
+                6
+            ]
+        ]
+    },
+    "backbone": "$monai.networks.nets.resnet.resnet50(spatial_dims=3,n_input_channels=1,conv1_t_stride=[2,2,1],conv1_t_size=[7,7,7])",
+    "feature_extractor": "$monai.apps.detection.networks.retinanet_network.resnet_fpn_feature_extractor(@backbone,3,False,[1,2],None)",
```
Nic-Ma, line 54: If these 2 components are class instances, I think we should define with `_target_` args?
@yiheng-wang-nv What do you think?

Thanks.

Can-Zhao: Thanks Nic. They are not class instances. They are functions.

Can-Zhao:
> Hi @Can-Zhao , the schema check is failed (https://github.com/Project-MONAI/model-zoo/runs/7832105295?check_suite_focus=true#step:2:219), according to the schema this bundle used, the following properties are required:
> 
> ```
> "type",
> "format",
> "num_channels",
> "spatial_shape",
> "dtype",
> "value_range"
> ```

Thank you! It is resolved

models/lung_nodule_ct_detection/configs/inference.json
```diff
@@ -0,0 +1,200 @@
...
+        "_target_": "SlidingWindowInferer",
+        "roi_size": [
+            240,
+            240,
+            160
+        ],
+        "sw_batch_size": 1,
+        "overlap": 0.5
+    },
+    "output_transform": {
```
MMelQin: Should the attribute name be the more common/agreed on keyword, `postprocessing`, like `preprocessing`? In any case, I inspected this bundle and used the built-in name mapping mechanism to work around this.

models/lung_nodule_ct_detection/configs/inference.json
```diff
@@ -0,0 +1,200 @@
...
+        "debug": false
+    },
+    "detector_ops": [
+        "$@detector.set_target_keys(box_key='box', label_key='label')",
+        "$@detector.set_box_selector_parameters(score_thresh=0.02,topk_candidates_per_level=1000,nms_thresh=0.22,detections_per_img=300)",
+        "$@detector.set_sliding_window_inferer(roi_size=@val_patch_size,overlap=0.25,sw_batch_size=1,mode='constant',device='cpu')"
+    ],
+    "preprocessing": {
+        "_target_": "Compose",
+        "transforms": [
```
MMelQin, line 72: In real world use in an inference application, the original DICOM CT input will be parsed to find the specific series to extract a single (numpy) image, along with metadata, e.g. spacing. The spacing can vary somewhat, so in the typical inference config JSON, there needs to be a SpacingD transform to resample the image. Typically model trainer ignore this step as they process the cohort of images before hand.

models/lung_nodule_ct_detection/configs/metadata.json
```diff
@@ -0,0 +1,72 @@
...
+                "format": "magnitude",
+                "modality": "CT",
+                "num_channels": 1,
+                "spatial_shape": [
+                    "16*n",
+                    "16*n",
+                    "8*n"
+                ],
+                "dtype": "float16",
+                "value_range": [
```
MMelQin, line 44: If this referring to the already processed data from a CT DICOM series? Original CT image instance/slice is typically 512x512 (16 bit INT per pixel) and N slices (not necessarily 8*n). Also, the image is typically grayscale, Monochrome2 mostly, so the value range would not be [0, 1].

By reading this and the inference config, it looks to me that the original CT image needs to be resampled to the specific spacing and whatever pixel counts. During sliding window inference, the patch is then padded to be of the expected (16*n, 16*n, 8*m) shape, and in fact (240, 240, 160) per inferer ROI. This is common. But the README.md states `Actual Model Input: 192 x 192 x 80`, different from the inferer ROI.

models/lung_nodule_ct_detection/configs/inference.json
Can-Zhao: I will fix it

models/lung_nodule_ct_detection/configs/inference.json
Can-Zhao: I will add it. Thanks!

models/lung_nodule_ct_detection/configs/metadata.json
Can-Zhao: This refers to the network input. The network is part of the detector. The detector will pad the image to (16n, 16n, 8*m), so the detector can take arbitrary image shape.

models/lung_nodule_ct_detection/configs/inference.json
Can-Zhao: Added!

Nic-Ma:
Hi @MMelQin ,

Do you have any other comments for this PR?

Thanks in advance.

models/lung_nodule_ct_detection/configs/metadata.json
```diff
@@ -0,0 +1,72 @@
+{
+    "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
+    "version": "0.1.0",
+    "changelog": {
+        "0.1.0": "complete the model package"
+    },
+    "monai_version": "0.9.0",
```
Nic-Ma: Hi @Can-Zhao ,

Could you please help update to test and use the MONAI 0.9.1?
We would like to support the latest version in this important bundle.

Thanks in advance.

models/lung_nodule_ct_detection/configs/metadata.json
```diff
@@ -0,0 +1,72 @@
...
+    "pytorch_version": "1.10.0",
```
Nic-Ma: And the latest PyTorch version.

Thanks.

models/lung_nodule_ct_detection/scripts/cocometric_ignite.py
```diff
@@ -0,0 +1,112 @@
...
+import torch
+from ignite.metrics.metric import Metric, reinit__is_reduced, sync_all_reduce
+from monai.apps.detection.metrics.coco import COCOMetric
+from monai.apps.detection.metrics.matching import matching_batch
+from monai.data import box_utils
+
+from .utils import detach_to_numpy
+
+
+class IgniteCocoMetric(Metric):
```
Nic-Ma, line 12: Is it possible to inherit this `IgniteMetric` class to simplify the logic?
https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/ignite_metric.py#L32
I think you can refer to the `DiceMetrc`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/handlers/mean_dice.py#L19
And actually, I feel maybe you can move this handler to MONAI apps directly?
CC @wyli What do you think?

Thanks.

models/lung_nodule_ct_detection/scripts/warmup_scheduler.py
```diff
@@ -0,0 +1,88 @@
...
+
+"""
+This script is adapted from
+https://github.com/ildoonet/pytorch-gradual-warmup-lr/blob/master/warmup_scheduler/scheduler.py
+"""
+
+from torch.optim.lr_scheduler import ReduceLROnPlateau, _LRScheduler
+
+
+class GradualWarmupScheduler(_LRScheduler):
```
Nic-Ma, line 20: Is it possible to move this component to MONAI core?
Seems like a common and useful LrScheduler.

Thanks.

Nic-Ma (commented):
Hi @Can-Zhao ,

Thanks for putting together the detection bundle.
It overall looks good to me, put some minor comments inline.
I feel maybe it contains too much python code, better to move some into MONAI.
@wyli @yiheng-wang-nv Could you please help also review it?

Thanks in advance.

models/lung_nodule_ct_detection/scripts/cocometric_ignite.py
Can-Zhao: Inheriting from IgniteMetric cannot simply it...It follows a different way of aggregation from most other metrics

models/lung_nodule_ct_detection/scripts/warmup_scheduler.py
Can-Zhao: Last time I had a discussion with @wyli , and decided to leave it here for now...

Nic-Ma (approved):

PR #116, opened 2022-09-09
52 Unify bundle naming to support other apps

### LINKED ISSUE #52: Resolve conflicting key names across bundles
Conflict key names:
https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_deepedit_annotation/configs/inference.json#L63

vs

https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_ct_segmentation/configs/inference.json#L33

Is there a validation script to check against all configs.. or user can choose any name in such cases?

yiheng-wang-nv:
Signed-off-by: yiheng-wang-nv <vennw@nvidia.com>

Fixes #52 .

### Description
Progress:
1. change naming format for all related bundles.
2. change `ci/run_premerge_gpu.sh` in order to prevent install environment conflicts when multiple bundles are changed.
3. change `ci/verify_bundle.py`, and add key name requirements to unify bundle naming.

### Status
**Ready**


CONTRIBUTING.md
```diff
@@ -56,13 +56,32 @@ large_files:
     url: "url-of-model.ts"
 ```
 
+### Necessary Files and Keys
+
+In order to be compatible with other apps such as MONAI FL, MONAI deploy and MONAI Label, except the MONAI Bundle Specification, a bundle in MONAI Model Zoo should contain the following necessary files:
```
Nic-Ma: I think we should describe "for the typical supervised segmentation and classification tasks...". Other tasks no need to follow the naming so far and maybe can't follow at all.
@wyli What do you think?

Thanks.

yiheng-wang-nv: HI @Nic-Ma , only these files are hard requirements:
```
1. `configs/metadata.json`
2. `models/model.pt` (or a download link in the config file for large files)
3. `configs/inference.json` (or `.yaml`, `.json`)
```
It seems these files are necessary (correct me if I made a mistake here)

Other requirements all have prerequisites (like if using preprocessing, it should be named as `preprocessing` exactly), and I think that using unify naming rules no matter which task the bundle belongs to is helpful if we have further supports on new tasks, for example, support GAN in MONAI deploy. Therefore, users do not need to worry about the situation that their non typical bundles cannot follow the rules. What do you think?

wyli: maybe `Necessary Files and Keys` should be `Preferred Files and Keys`, so that the bundle can have good interoperability with other tools such as MONAI FL, deploy SDK and Label...

ci/bundle_custom_data.py
```diff
@@ -15,6 +15,10 @@
 # If a bundle does not need to be tested, please add the bundle name into the list.
 exclude_verify_shape_list = ["mednist_gan", "lung_nodule_ct_detection"]
 
+# This list is used for our CI tests to determine whether a bundle contains the preferred files.
+# If a bundle does not have any of the preferred files, please add the bundle name into the list.
+exclude_verify_preferred_files_list = []
```
Nic-Ma, line 20: maybe "preferred_files_or_keys"?

Thanks.

Nic-Ma (approved):

Nic-Ma:
Hi @wyli ,

Could you please help also review the PR?

Thanks in advance.

CONTRIBUTING.md
```diff
@@ -56,13 +56,35 @@ large_files:
     url: "url-of-model.ts"
 ```
 
+### Preferred Files and Keys
+
+In order to be compatible with other apps such as MONAI FL, MONAI deploy and MONAI Label, except the MONAI Bundle Specification, a bundle in MONAI Model Zoo should contain the necessary file `configs/metadata.json` and the following preferred files:
```
wyli: perhaps put a link of example here explaining the definition of the files and keys?

wyli (approved):

ci/bundle_custom_data.py
yiheng-wang-nv: Hi @Nic-Ma , requirements of keys are not hard requirements, thus we may not need to add them here. For example, if a bundle does not need to use `postprocessing`, our ci test will skip the check of it directly. Otherwise, we need to maintain a dict for multiple keys and bundle developer need to classify them one by one.

CONTRIBUTING.md
yiheng-wang-nv: Thanks @wyli , examples are added.

ci/bundle_custom_data.py
Nic-Ma: OK, got it.

Thanks.

PR #133, opened 2022-09-20
Updating configuration scripts and readme for bundle

ericspod:
Signed-off-by: Eric Kerfoot <eric.kerfoot@kcl.ac.uk>

### Description
Updates to the scripts for 1.0 and readme file. Moved common definitions to a common file used for both inference and training.

### Status
**Ready**


ericspod:
The issue appears to be the fact that two config files are needed when running training or inference. There's a `common.json` file that is needed by `train.json` or `inference.json`.

models/valve_landmarks/configs/common.json
```diff
@@ -0,0 +1,40 @@
+{
```
yiheng-wang-nv: Hi @ericspod , in MONAI label, deploy and FL, it is required to have some necessary keys within the train.json and inference.json, for example: `network_def`, `bundle_root` (like in: https://github.com/Project-MONAI/MONAILabel/blob/main/sample-apps/monaibundle/lib/trainers/bundle.py#L30). You can also refer to: https://github.com/Project-MONAI/model-zoo/issues/52 for more details.
These requirements are for bundles in model zoo, thus could we go back to put these common components back? Or do you have any other suggestions? Thanks!

ericspod:
I've just changed the scripts back to how they were.

models/valve_landmarks/configs/common.json
wyli: I think maybe we should allow for different versions of the same bundle item, with some versions support the preferred keywords and the others less restrictive?

wyli (approved):

models/valve_landmarks/configs/common.json
ericspod: It wasn't an issue of keywords being absent, the issue was that multiple config files were needed together, ie. this was in the command line: `-config_file "['configs/inference.json', 'configs/common.json']"` The `common.json` file has common definitions, this eliminated duplication but had the side effect of not testing correctly. I'll propose a feature later but we need some way of including one config file in another with a `_include_` tag or something like that.

Nic-Ma: Hi @ericspod ,

Our `%` syntax supports macro replacement, so actually we can use `"include": "%my/another/config.json"`?

Thanks.

ericspod: Would it recognise that as a filename and expand all the items in that file even without any # syntax? Is the path relative to the current file? If yes that would be the solution though we should discuss it in more detail in the docs.

Nic-Ma: @ericspod yes, but the path may be same as the CLI. For more details, please take a look at: https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/config_parser.py#L327

Thanks.

ericspod: Thanks Nic, this will work but one would have to refer to components by the key for the macro, eg. "$@include.network_def.to(@include.device)". If might be easier to have a `_include_` tag that would copy values into the current namespace and also give feedback if a circular import is encountered.

Nic-Ma: Interesting, so it's something like `macro replacement` not only on values with `%`, but also on the keys.
CC @wyli @yiheng-wang-nv for more discussion.

Thanks.

PR #143, opened 2022-09-29
Add training pipeline in renal and whole brain seg bundle to support active model fine-tuning in monailabel

tangy5:
Asked by a user from monailabel side, we add training pipeline for the renal seg and whole brain seg bunlde so that users can use active learning, fine-tuned labels to train models. 

1. Training experiments conducted and tested with bundle config files.
2. Training pipeline, tuned hyper-parameter and transformations are set to renal and whole brain segmentation.
3. Modified metadata.josn and readme files.

Training tested.


yiheng-wang-nv (approved):

yiheng-wang-nv:
Thanks @tangy5 , this PR looks good to me.

PR #163, opened 2022-10-14
156 split loader and saver from processing

yiheng-wang-nv:
Fixes #156 .

### Description
This PR modifies the inference script of `spleen_ct_segmentation` bundle via putting image load out of `preprocessing` and putting image save out of `postprocessing.`
a flag `use_loader` is used to determine if `LoadImaged` is included into `preprocessing`, and a flag `use_saver` is used to determine if `SaveImaged` is included into `postprocessing`.

After these changes:
- if users want to use the default settings, they do not need to change anything.
- if users want to load images by using other methods, but still use the `dataset` inside the script, they can specify `use_loader` to `False` and then override `dataset#data`, the following is an example:

Screen Shot 2022-10-14 at 4 01 14 PM

### Status
**Ready**

### Please ensure all the checkboxes:

- [x] Codeformat tests passed locally by running `./runtests.sh --codeformat`.
- [ ] In-line docstrings updated.
- [ ] Update `version` and `changelog` in `metadata.json` if changing an existing bundle.
- [ ] Please ensure the naming rules in config files meet our requirements (please refer to: `CONTRIBUTING.md`).
- [ ] Ensure versions of packages such as `monai`, `pytorch` and `numpy` are correct in `metadata.json`.
- [ ] Descriptions should be consistent with the content, such as `eval_metrics` of the provided weights and TorchScript modules.
- [ ] Files larger than 25MB are excluded and replaced by providing download links in `large_file.yml`.
- [ ] Avoid using path that contains personal information within config files (such as use `/home/your_name/` for `"bundle_root"`).

PR #167, opened 2022-10-18
add inference config for raw images in detection module, update readme

### LINKED ISSUE #166: Lung Nodule detection (Inference)
I have followed the instructions to do the inference but I have three questions: 
1- The output I got having three main things: 
- the ground truth coordinates
- label 
- image name
I’m wondering why the label is always zero? but some images have two zeros and some have more? can you please elaborate more on this part?

2- The inference is working only on the raw data (The unprocessed images), while when I infer the model on the resampled data it is not working!?

3- The inference uses the ground truth and calculates the label score, what if I want to do the inference on a different dataset that has no ground truth? 
I want to see the performance of the model on another dataset.

Can-Zhao:

Fixes #166 .

### Description
Add inference config file for raw input images.
Update readme to add more explanation on inference

### Status


Nic-Ma:
Hi @Can-Zhao ,

To avoid duplicated config content, I would suggest to just define `inference_raw.json` to override `inference.json`. Like your `evaluate.json` to override the `train.json`:
https://github.com/Project-MONAI/model-zoo/blob/dev/models/lung_nodule_ct_detection/configs/evaluate.json
Or maybe you can just add a flag in the `inference.json` to control the raw or not-raw, refer to:
https://github.com/Project-MONAI/model-zoo/pull/163
CC @yiheng-wang-nv 

Thanks.

Can-Zhao:
> Hi @Can-Zhao ,
> 
> To avoid duplicated config content, I would suggest to just define `inference_raw.json` to override `inference.json`. Like your `evaluate.json` to override the `train.json`: https://github.com/Project-MONAI/model-zoo/blob/dev/models/lung_nodule_ct_detection/configs/evaluate.json Or maybe you can just add a flag in the `inference.json` to control the raw or not-raw, refer to: #163 CC @yiheng-wang-nv
> 
> Thanks.

Thank you! I make it override the inference.json

yiheng-wang-nv (commented):
Hi @Can-Zhao , please also modify the `version` and `changelog`, thanks!

models/lung_nodule_ct_detection/configs/inference_raw.json
```diff
@@ -0,0 +1,55 @@
+{
+    "data_list_file_path": "$@bundle_root + '/LUNA16_datasplit/original/dataset_fold0.json'",
+    "data_file_base_dir": "/home/canz/Projects/datasets/LUNA16/93176/Images",
+    "preprocessing": {
+        "_target_": "Compose",
+        "transforms": [
+            {
+                "_target_": "DeleteItemsd",
```
Nic-Ma: May I know the error or problem if don't remove these 2 keys?

Thanks.

models/lung_nodule_ct_detection/configs/inference_raw.json
```diff
@@ -0,0 +1,55 @@
...
+                "keys": [
+                    "box",
+                    "label"
+                ]
+            },
+            {
+                "_target_": "LoadImaged",
+                "keys": "image",
+                "meta_key_postfix": "meta_dict",
+                "reader": "itkreader",
```
Nic-Ma: I feel the only difference in this new config is this line, no need to override all the transforms.
Maybe you can just add a flag in the original `inference.json`? For example:
```json
"resampled_image": "true",
...
"reader": "$'nibabelreader' if @resampled_image else 'itkreader'",
...
```

What do you think?
CC @yiheng-wang-nv @wyli .

Thanks.

yiheng-wang-nv: Agree. Similar for the `Spacingd` transform, can add a flag and make use of `_disabled_` to control whether to use this transform. Can refer to [here](https://github.com/Project-MONAI/model-zoo/pull/163/files#diff-53004b9136c53f2fc7834c750c784abc55fc3161cfe1c53c1143dde5ae581a23R127)

models/lung_nodule_ct_detection/configs/inference_raw.json
```diff
@@ -0,0 +1,55 @@
+{
+    "data_list_file_path": "$@bundle_root + '/LUNA16_datasplit/original/dataset_fold0.json'",
```
Can-Zhao: If we use `"_disabled_"`, how to handle the change of `"data_list_file_path"` of the two setting? @Nic-Ma @yiheng-wang-nv . Sorry, I'm still not that familiar with the commands of bundle

models/lung_nodule_ct_detection/configs/inference_raw.json
Can-Zhao: There is no error resulted. I have removed this transform. I added only to make sure the inference did not use ground truth.

models/lung_nodule_ct_detection/configs/inference_raw.json
Can-Zhao: There are at least three places to switch. I was afraid it is not that convenient to use `"_disabled_"`

models/lung_nodule_ct_detection/configs/inference_raw.json
Nic-Ma: Actually, I feel we don't need to care about the `data_list_file_path`, if it's raw image, you need to set `resampled_image=True`.
Don't worry about the bundle APIs, I think @yiheng-wang-nv can help you on the development.

Thanks.

models/lung_nodule_ct_detection/configs/inference_raw.json
Nic-Ma: Hi @yiheng-wang-nv , I don't see any difference in the `Spacingd` transform, what do you mean to change for `resampled or raw`?

Thanks.

models/lung_nodule_ct_detection/configs/inference_raw.json
Can-Zhao: But here, we provided download link for all the data list json files. I put json files for resampled and raw images in two different folders in the downloaded file. It is already structured. That's why I put a `"data_list_file_path"` here.

models/lung_nodule_ct_detection/configs/inference_raw.json
Can-Zhao: There is a change. When apply to resampled image, we should not include spacing

models/lung_nodule_ct_detection/configs/inference.json
```diff
@@ -71,16 +73,18 @@
...
             {
-                "_target_": "DeleteItemsd",
-                "keys": [
-                    "box",
-                    "label"
-                ]
+                "_target_": "LoadImaged",
+                "keys": "image",
+                "meta_key_postfix": "meta_dict",
+                "_disabled_": "@whether_raw_luna16"
```
yiheng-wang-nv, line 79: Hi @Can-Zhao , thanks for the updates. Here I think we may not need two `LoadImaged `, and you can use:
```
            {
                "_target_": "LoadImaged",
                "keys": "image",
                "meta_key_postfix": "meta_dict",
                "reader": "$'itkreader' if @whether_raw_luna16 is True else None",
                "affine_lps_to_ras": "@whether_raw_luna16",
            },
```
In addition, according to the readme file, should the `affine_lps_to_ras` of `AffineBoxToWorldCoordinated` be consistent with this transform?

Can-Zhao: I'm afraid they cannot be combined.
1) the resampled images are nifti. nifti reader does not take `affine_lps_to_ras` as an argument, it will give an error.
2) that's why I put two `"LoadImaged"` here.

yiheng-wang-nv: Thanks @Can-Zhao , I double checked the settings with nii format data and you are right.
Could you please also update the changelog and version in metadata as I mentioned in a previous comment? It is a necessary step before merging a PR, thanks again!

Can-Zhao: Thank you! Just changed them!

yiheng-wang-nv (approved):

PR #245, opened 2022-11-22
Build Pathology HoVerNet Model Zoo Bundle

### LINKED ISSUE #5181: Pathology HoVerNet: Model Zoo Bundle

KumoLiu:
Fixes https://github.com/Project-MONAI/MONAI/issues/5181

### Description
Build HoVerNet model zoo bundle

### Status

- [x] In-line docstrings updated.


models/pathology_nuclei_segmentation_classification/docs/README.md
```diff
@@ -0,0 +1,124 @@
+# Model Overview
+A pre-trained model for simultaneous segmentation and classification of nuclei within multitissue histology images based on CoNSeP data. The details of the model can be found in [1]
+
+## Workflow
+
+The model is trained to simultaneous segment and classify nuclei. Training is done via a two-stage approach. First initialised the model with pre-trained weights on the [ImageNet dataset](https://ieeexplore.ieee.org/document/5206848), trained only the decoders for the first 50 epochs, and then fine-tuned all layers for another 50 epochs.
+
+- Each user is responsible for checking the content of models/datasets and the applicable licenses and determining if suitable for the intended use.The license for the pre-trained model used in examples is different than MONAI license. Please check the source where these weights are obtained from: https://github.com/vqdang/hover_net#data-format
+
+pretrained_model = "https://drive.google.com/u/1/uc?id=1KntZge40tAHgyXmHYVqZZ5d2p_4Qr2l5&export=download"
```
yiheng-wang-nv: HI @KumoLiu , I suggest to rewrite this line since it should be a sentence rather than a line of python code in readme, thanks!

models/pathology_nuclei_segmentation_classification/docs/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+The achieved metrics on the validation data are:
+- Binary Dice: 0.82762
+- PQ: 0.48976
+- F1d: 0.73592
+
+#### Training Loss and Dice
+
+stage1:
+![A graph showing the training loss and the mean dice over 50 epochs in stage1](https://drive.google.com/file/d/1sc0ZqKWmphZbhHxKu-UredYriFtUFlbk/view?usp=sharing)
```
yiheng-wang-nv: It seems this way cannot produce a figure in the readme file.
see: https://github.com/Project-MONAI/model-zoo/blob/d8317ab6badeadb3ed94955db2d6d5f22aa6ad3e/models/pathology_nuclei_segmentation_classification/docs/README.md#training-loss-and-dice

models/pathology_nuclei_segmentation_classification/docs/README.md
```diff
@@ -0,0 +1,124 @@
...
+- AMP: True
+- Optimizer: Adam
+- Learning Rate: 1e-4
+- Loss: HoVerNetLoss
+
+## Input
+
+Input: RGB images
+
+1. Randomly performing rotate, scale, shear, translate, etc
```
yiheng-wang-nv: I'm not sure but it seems in the readme file, there is no need to describe the preprocessing and augmentation methods (they are already in configs). During @zephyrie 's last round of review, similar descriptions in other bundles are removed (see: https://github.com/Project-MONAI/model-zoo/pull/233/files#diff-21a8c44891c2467d971a975c73d4488b87963f0366b003847fea6d959c4521adL43)

Hi @Nic-Ma , could you please help to double confirm it? Thanks!

models/pathology_nuclei_segmentation_classification/docs/README.md
```diff
@@ -0,0 +1,124 @@
...
+2. Cropping images to (270, 270)
+3. Randomly spatial flipping
+4. Randomly choosing one transform to smooth images
+5. Randomly adjusting brightness, saturation, contrast and hue
+6. Scaleing intensity to [0, 1]
+7. Generating horizontal and vertical distance maps from label instance map
+
+## Output
+
+Output: a dictionary
```
yiheng-wang-nv: a dictionary with the following keys:
1. nucleus_prediction: ...
2. ...
3. ...

models/pathology_nuclei_segmentation_classification/docs/README.md
KumoLiu: Thanks, @yiheng-wang-nv, I have rewritten it in https://github.com/Project-MONAI/model-zoo/pull/245/commits/ec8b93437c931018287a9b4b8e6a6d0670901b1f.

models/pathology_nuclei_segmentation_classification/docs/README.md
KumoLiu: Yes, I will update it once I get the correct links.

models/pathology_nuclei_segmentation_classification/docs/README.md
KumoLiu: Have updated it, thanks!

models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py
```diff
@@ -0,0 +1,219 @@
...
+import os
+import pathlib
+import shutil
+from argparse import ArgumentParser
+
+import cv2
+import numpy as np
+import scipy.io as sio
+import tqdm
+
```
yiheng-wang-nv, line 12: suggest to add cv2, scipy, tqdm into `metadata.json`

models/pathology_nuclei_segmentation_classification/configs/train.json
```diff
@@ -0,0 +1,502 @@
+{
+    "imports": [
+        "$import glob",
+        "$import os",
+        "$import skimage"
+    ],
+    "bundle_root": "/workspace/Code/model-zoo/models/pathology_nuclei_segmentation_classification",
```
yiheng-wang-nv: here can the `bundle_root` be `$os.getcwd()` as @holgerroth suggested in https://github.com/Project-MONAI/model-zoo/issues/247 ? Thanks!

models/pathology_nuclei_segmentation_classification/docs/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+- Each user is responsible for checking the content of models/datasets and the applicable licenses and determining if suitable for the intended use.The license for the pre-trained model used in examples is different than MONAI license. Please check the source where these weights are obtained from: https://github.com/vqdang/hover_net#data-format
+
+`pretrained_model` is "https://drive.google.com/u/1/uc?id=1KntZge40tAHgyXmHYVqZZ5d2p_4Qr2l5&export=download" which will be used in bash code below.
+
+![Model workflow](https://ars.els-cdn.com/content/image/1-s2.0-S1361841519301045-fx1_lrg.jpg)
+
+## Data
+
+The training data is from https://warwick.ac.uk/fac/cross_fac/tia/data/hovernet/.
```
yiheng-wang-nv: we need a `data_license.txt` in `docs`. Please take: https://github.com/Project-MONAI/model-zoo/blob/dev/models/pathology_tumor_detection/docs/data_license.txt for reference.

KumoLiu: The license of CoNSeP dataset is Apache2.0, the same as the bundle, do we still need to add the `data_license.txt` file?

yiheng-wang-nv: I think it is needed since apache 2.0 is also a license. You can do sth like: https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_ct_segmentation/docs/data_license.txt

models/pathology_nuclei_segmentation_classification/configs/train.json
KumoLiu: Have updated it in https://github.com/Project-MONAI/model-zoo/pull/245/commits/09a2938f73b2efed915f6e141bd262144e85d944, thanks!

bhashemian:
Hi @yiheng-wang-nv, do you know what this is about? 
```
11    B905 `zip()` without an explicit `strict=` parameter.
[82](https://github.com/Project-MONAI/model-zoo/actions/runs/3641203726/jobs/6146857366#step:7:83)
11
[83](https://github.com/Project-MONAI/model-zoo/actions/runs/3641203726/jobs/6146857366#step:7:84)
Check failed!
```
https://github.com/Project-MONAI/model-zoo/actions/runs/3641203726/jobs/6146857366#step:7:84

models/pathology_nuclei_segmentation_classification/docs/README.md
bhashemian: @Nic-Ma can you clarify this with @KumoLiu please? Thanks

models/pathology_nuclei_segmentation_classification/configs/inference.json
```diff
@@ -0,0 +1,150 @@
+{
+    "imports": [
+        "$import glob",
+        "$import os"
+    ],
+    "bundle_root": ".",
+    "output_dir": "$os.path.join(@bundle_root, 'eval')",
+    "dataset_dir": "/workspace/Data/CoNSeP/Test/Images",
```
bhashemian:
```suggestion
    "dataset_dir": "/workspace/Data/Pathology/CoNSeP/Test/Images",
```

models/pathology_nuclei_segmentation_classification/configs/train.json
```diff
@@ -0,0 +1,502 @@
...
+        "$import glob",
+        "$import os",
+        "$import skimage"
+    ],
+    "bundle_root": "$os.getcwd()",
+    "ckpt_dir_stage0": "$@bundle_root + '/models/stage0'",
+    "ckpt_dir_stage1": "$@bundle_root + '/models/stage1'",
+    "ckpt_path_stage0": "$@ckpt_dir_stage0 + '/checkpoint_epoch=50.pt'",
+    "output_dir": "$@bundle_root + '/eval'",
+    "dataset_dir": "/workspace/Data/CoNSeP/Prepared/",
```
bhashemian:
```suggestion
    "dataset_dir": "/workspace/Data/Pathology/CoNSeP/Prepared/",
```

models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py
```diff
@@ -0,0 +1,219 @@
...
+        pbarx.close()
+
+
+def parse_arguments():
+    parser = ArgumentParser(description="Extract patches from the original images")
+
+    parser.add_argument(
+        "--root",
+        type=str,
+        default="/home/yunliu/Workspace/Data/CoNSeP",
```
bhashemian:
```suggestion
        default="/workspace/Data/Pathology/CoNSeP",
```

models/pathology_nuclei_segmentation_classification/docs/README.md
Nic-Ma: @KumoLiu Please update according to the comments, thanks.

models/pathology_nuclei_segmentation_classification/configs/inference.json
```diff
@@ -0,0 +1,150 @@
...
+        "pin_memory": true
+    },
+    "inferer": {
+        "_target_": "SlidingWindowHoVerNetInferer",
+        "roi_size": "@patch_size",
+        "sw_batch_size": "@sw_batch_size",
+        "overlap": "$1.0 - float(@out_size) / float(@patch_size)",
+        "padding_mode": "constant",
+        "cval": 0,
+        "sw_device": "@device",
```
SachidanandAlle: remove sw_device and device fields and let it be None.. so that during inferer input based device gets selected..
similar strategy is used in other bundles like speen

models/pathology_nuclei_segmentation_classification/large_files.yml
```diff
@@ -0,0 +1,3 @@
+large_files:
+  - path: "models/model.pt"
+    url: "https://drive.google.com/file/d/1Auwm8YrQCx9EAX_oViNTlJ_JgBb-wbzZ/view?usp=sharing"
```
SachidanandAlle: this is wrong file.. it has dict {"net": } all others are using {"model": } as part of state dictionary..

models/pathology_nuclei_segmentation_classification/configs/inference.json
```diff
@@ -0,0 +1,150 @@
...
+            }
+        ]
+    },
+    "handlers": [
+        {
+            "_target_": "CheckpointLoader",
+            "load_path": "$os.path.join(@bundle_root, 'models', 'model.pt')",
+            "map_location": "@device",
+            "load_dict": {
+                "model": "@network"
```
SachidanandAlle, line 133: the large file used doesn't have "model" as load dict..
before raising the PR.. please make sure basic testing is done for infer/train workflows

models/pathology_nuclei_segmentation_classification/configs/inference.json
KumoLiu: Have updated, thx!

models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py
KumoLiu: Have updated, thx!

models/pathology_nuclei_segmentation_classification/configs/train.json
KumoLiu: Have updated, thx!

models/pathology_nuclei_segmentation_classification/docs/README.md
KumoLiu: Have removed it, thanks!

models/pathology_nuclei_segmentation_classification/docs/README.md
KumoLiu: Have updated it, thanks!

KumoLiu:
Hi @drbeh, I try to run the inference part in my local env, but get this error [error.log](https://github.com/Project-MONAI/model-zoo/files/10182910/error.log), could you please help take a look at it? And also help address comments by @SachidanandAlle in the inference part. Thanks!

models/pathology_nuclei_segmentation_classification/configs/inference.json
bhashemian: In all the bundles if gpu available, `@device` will b set to 'cuda:0' and the model will reside there, so why the input sliding window data (`sw_device`) should reside somewhere else?

bhashemian: Although removing `device` makes sense to me because it is about the output, and our post-processing is currently on cpu.

SachidanandAlle: if device is None.. doesn't mean it's happening on CPU..
https://github.com/Project-MONAI/MONAI/blob/dev/monai/inferers/utils.py#L102-L108
https://github.com/Project-MONAI/MONAI/blob/dev/monai/inferers/utils.py#L129-L132

if input is already loaded into cuda.. things will happen accordingly
same for sw_device

SachidanandAlle: if you have observed when input in cuda but things happening on cpu.. then it's a bug.. u might want to create an issue the same on monai github

SachidanandAlle: so i would rather keep default based on the input rather force the device type.. even for deploy they only consider the transforms/inferers at the end.. and currently bundle api doesn't support loading particular section with overrides (to override the GPU selected by user)

I am referring `get_parsed_content(k, instantiate=True)` from monai bundle apis.. it doesnt support passing any overrides for the attributes.

models/pathology_nuclei_segmentation_classification/configs/train.json
```diff
@@ -0,0 +1,502 @@
+{
+    "imports": [
+        "$import glob",
+        "$import os",
+        "$import skimage"
+    ],
+    "bundle_root": "$os.getcwd()",
+    "ckpt_dir_stage0": "$@bundle_root + '/models/stage0'",
+    "ckpt_dir_stage1": "$@bundle_root + '/models/stage1'",
+    "ckpt_path_stage0": "$@ckpt_dir_stage0 + '/checkpoint_epoch=50.pt'",
```
SachidanandAlle: why such hardcoding?
why can't we follow the similar standard by referring: 
https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_ct_segmentation/configs/train.json
https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_ct_segmentation/configs/train.json#L255


```
    "ckpt_dir": "$@bundle_root + '/models'",
    "output_dir": "$@bundle_root + '/eval'",
```

as per the bundle spec.. you try to produce model.pt (which can serve the current best model as part of your training pipeline)...

i strongly suggest to read through the existing (spleen) example and fix some of these things..

models/pathology_nuclei_segmentation_classification/configs/train.json
```diff
@@ -0,0 +1,502 @@
...
+                "_target_": "ValidationHandler",
+                "validator": "@validate#evaluator",
+                "epoch_level": true,
+                "interval": "@val_interval"
+            },
+            {
+                "_target_": "CheckpointSaver",
+                "save_dir": "@ckpt_dir",
+                "save_dict": {
+                    "net": "@network",
```
SachidanandAlle: you are producing `net` inferer is using `model`.. both needs to be in sync at the end of the day.. and let them follow again similar names/standards like others while saving the checkpoint

models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py
```diff
@@ -0,0 +1,219 @@
...
+
+            # *
+            img = np.concatenate([img, ann], axis=-1)
+            sub_patches = xtractor.extract(img, cfg["extract_type"])
+
+            pbar_format = "Extracting  : |{bar}| {n_fmt}/{total_fmt}[{elapsed}<{remaining},{rate_fmt}]"
+            pbar = tqdm.tqdm(total=len(sub_patches), leave=False, bar_format=pbar_format, ascii=True, position=1)
+
+            for idx, patch in enumerate(sub_patches):
+                image_patch = patch[..., :3].transpose(2, 0, 1)  # make channel first
```
SachidanandAlle: some of these should go as part of training pre transforms.. like EnsureChannelFirst, Adding Channel for label...
This should only produce the required image/label/type_map files which can be either numpy or png and hence loadable in LoadImage transform.

This will enable others to prepare the similar dataset for additional training.. u have to separate and simplify things wrt ConSeP data and non ConSeP dataset..

models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py
```diff
@@ -0,0 +1,219 @@
...
+                np.save("{0}/{1}_{2:03d}_type_map.npy".format(out_dir, base_name, idx), type_map_patch)
+                pbar.update()
+            pbar.close()
+            # *
+
+            pbarx.update()
+        pbarx.close()
+
+
+def parse_arguments():
```
SachidanandAlle, line 201: I have fewer 2000x2000 newly annotated image + label files.. can i use this utility to prepare the required patches to train or fine tune the model?  what is the requirement to create new patches for a given image+label pair?

label => labelmap (mask with class values [0...4])

``` json
{
  "Background": 0,
  "Other": 1,
  "Inflammatory": 2,
  "Epithelial": 3,
  "Spindle-Shaped": 4
}
```

models/pathology_nuclei_segmentation_classification/configs/inference.json
bhashemian: I look a bit more into it and it seems that it doesn't matter to set `sw_device` to `None` or `@device` since we are setting `device` argument in `SupervisedEvaluator` to `@device`, which sends the input to the same `@device`.

models/pathology_nuclei_segmentation_classification/configs/train.json
bhashemian: @SachidanandAlle Maybe that is fine considering that is a two stage training unlike the other bundles. 
I personally think that it is better to have one standard bundle in a way that the user be able to run it with different input arguments for two-stage training. What do you think @KumoLiu?

models/pathology_nuclei_segmentation_classification/configs/train.json
bhashemian: @KumoLiu, I agree with @SachidanandAlle that we should be consistent with other bundles. But how long re-training would take?

models/pathology_nuclei_segmentation_classification/configs/inference.json
bhashemian: I removed it anyways.

models/pathology_nuclei_segmentation_classification/configs/train.json
bhashemian: Actually, we can just rename the key in the checkpoint, and run a training later to double check it.

models/pathology_nuclei_segmentation_classification/large_files.yml
```diff
@@ -0,0 +1,5 @@
+large_files:
+  - path: "models/model.pt"
+    url: "https://drive.google.com/uc?id=1Auwm8YrQCx9EAX_oViNTlJ_JgBb-wbzZ"
```
bhashemian: Can you update this model to have `model` key and re-upload?
```python
import torch
model = torch.load('model.pt')
model['model'] = model['net']
del model['net']
torch.save(model, 'model.pt')
```

Then can you also update the `hash_val` please? 
```
md5sum model.pt
```

models/pathology_nuclei_segmentation_classification/configs/train.json
```diff
@@ -0,0 +1,504 @@
...
+    "device": "$torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')",
+    "stage": 0,
+    "epochs": 50,
+    "val_interval": 1,
+    "learning_rate": 0.0001,
+    "amp": true,
+    "ckpt_dir": "$@ckpt_dir_stage0 if @stage == 0 else @ckpt_dir_stage1",
+    "network_def": {
+        "_target_": "HoVerNet",
+        "mode": "original",
```
bhashemian: Hi @KumoLiu 
Is there any reason that you are using `original` mode and not the `fast` mode?
Considering that the result of this training will be used in the inference, and inference time of `fast` mode is better, do you think that you can change all of them to `fast` mode? Thanks

models/pathology_nuclei_segmentation_classification/configs/train.json
SachidanandAlle: i am ok with staging.. my main point is `checkpoint_epoch=50.pt`

bhashemian: Oh yeah, agree. That is already fixed.

models/pathology_nuclei_segmentation_classification/configs/train.json
KumoLiu: Hi @drbeh, yes I'm ok to use any of them.

models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py
KumoLiu: Hi @SachidanandAlle, I think nothing special should be considered, only this `load_ann` function, which is specific to preparing the CoNSeP dataset. 
https://github.com/Project-MONAI/model-zoo/blob/3bff1f2de0ed7336eb627c16f04663105e39bde7/models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py#L18

And also this part, If you think we need to be compatible with some of the user's own data such as the data path, maybe I can make it a little more robust. 
https://github.com/Project-MONAI/model-zoo/blob/3bff1f2de0ed7336eb627c16f04663105e39bde7/models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py#L155-L174

models/pathology_nuclei_segmentation_classification/configs/train.json
bhashemian: Perfect, so it would be great if you can change it to "fast" across the bundle. Thanks

bhashemian: especially the weights. thanks

models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py
KumoLiu: Hi @SachidanandAlle, thanks for pointing it out, have removed it in this [commit](https://github.com/Project-MONAI/model-zoo/pull/245/commits/ff55e7e8333e1818b30e04ccc4a874f6d86fb95b).

models/pathology_nuclei_segmentation_classification/large_files.yml
KumoLiu: Have updated it, thanks!

models/pathology_nuclei_segmentation_classification/configs/train.json
KumoLiu: Have changed the mode to "fast" and also added fast mode metrics, thanks!

models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py
bhashemian: Hi @SachidanandAlle, what do you think about it?

SachidanandAlle: 👍

KumoLiu: Hi @SachidanandAlle, could you please help review it again, thanks!

models/pathology_nuclei_segmentation_classification/docs/README.md
```diff
@@ -4,7 +4,7 @@ A pre-trained model for simultaneous segmentation and classification of nuclei w
 
 ## Workflow
 
-The model is trained to simultaneous segment and classify nuclei. Training is done via a two-stage approach. First initialized the model with pre-trained weights on the [ImageNet dataset](https://ieeexplore.ieee.org/document/5206848), trained only the decoders for the first 50 epochs, and then fine-tuned all layers for another 50 epochs.
+The model is trained to simultaneous segment and classify nuclei. Training is done via a two-stage approach. First initialized the model with pre-trained weights on the [ImageNet dataset](https://ieeexplore.ieee.org/document/5206848), trained only the decoders for the first 50 epochs, and then fine-tuned all layers for another 50 epochs. There are two training modes in total. If "original" mode is specified, using [270, 270] and [80, 80] for `patch_size` and `out_size` respectively. If "fast" mode is specified, using [256, 256] and [164, 164] for `patch_size` and `out_size` respectively. The results we show below are based on the "fast" model.
```
bhashemian:
```suggestion
The model is trained to simultaneous segment and classify nuclei. Training is done via a two-stage approach. First initialized the model with pre-trained weights on the [ImageNet dataset](https://ieeexplore.ieee.org/document/5206848), trained only the decoders for the first 50 epochs, and then fine-tuned all layers for another 50 epochs. There are two training modes in total. If "original" mode is specified, it uses [270, 270] and [80, 80] for `patch_size` and `out_size` respectively. If "fast" mode is specified, it uses [256, 256] and [164, 164] for `patch_size` and `out_size` respectively. The results we show below are based on the "fast" model.
```

models/pathology_nuclei_segmentation_classification/docs/README.md
```diff
@@ -56,15 +56,10 @@ Output: a dictionary with the following keys:
...
-- PQ: 0.48976
-- F1d: 0.73592
-
 Fast mode:
-- Binary Dice: 0.83398
-- PQ: 0.50675
-- F1d: 0.73692
+- Binary Dice: 0.83285
+- PQ: 0.49765
+- F1d: 0.74214
```
bhashemian:
```suggestion
- Binary Dice: 0.8329
- PQ: 0.4977
- F1d: 0.7421
```

models/pathology_nuclei_segmentation_classification/scripts/prepare_patches.py
```diff
@@ -16,6 +16,10 @@ def load_img(path):
 
 
 def load_ann(path):
+    """
+    This function is specific to CoNSeP dataset.
+    if own dataset is used, then below may need to be modified
```
bhashemian:
```suggestion
    If using other datasets, the code below may need to be modified.
```

bhashemian (commented):
Hi @KumoLiu, it looks good to me. Just added few comments on the texts.

drbeh:
@KumoLiu this one needs to be fixed too. Thanks
https://github.com/Project-MONAI/tutorials/blob/5a497332b2b9df201d59e60f9f1062a11babc144/pathology/hovernet/inference.py#L136

KumoLiu:
Hi @drbeh, I have updated all the results, this PR is only waiting for the update of the performance curve from Michael.  Could you please help double-check the license descriptions in the README? If everything is ok, this PR can be merged first and I will create another PR to update the curve once I get the link. Thanks!

bhashemian:
Hi @KumoLiu, the inference is not working anymore because of the post-processing. Do you know why?

```
2022-12-13 21:09:05,835 - ignite.engine.engine.SupervisedEvaluator - ERROR - Current run is terminating due to exception: applying transform <monai.transforms.compose.Compose object at 0x7f3490071b20>
2022-12-13 21:09:05,862 - ignite.engine.engine.SupervisedEvaluator - ERROR - Engine run is terminating due to exception: applying transform <monai.transforms.compose.Compose object at 0x7f3490071b20>
Traceback (most recent call last):
  File "/home/bhashemian/workspace/MONAI/monai/transforms/transform.py", line 102, in apply_transform
    return _apply_transform(transform, data, unpack_items)
  File "/home/bhashemian/workspace/MONAI/monai/transforms/transform.py", line 66, in _apply_transform
    return transform(parameters)
  File "/home/bhashemian/workspace/MONAI/monai/apps/pathology/transforms/post/dictionary.py", line 535, in __call__
    d[self.instance_info_key], d[self.instance_map_key] = self.instance_map_post_process(
  File "/home/bhashemian/workspace/MONAI/monai/apps/pathology/transforms/post/array.py", line 735, in __call__
    instance_contour = self.generate_instance_contour(instance_mask, offset)
  File "/home/bhashemian/workspace/MONAI/monai/apps/pathology/transforms/post/array.py", line 561, in __call__
    inst_contour = generate_contour(inst_contour_cv)
  File "/home/bhashemian/workspace/MONAI/monai/apps/pathology/transforms/post/array.py", line 462, in __call__
    raise ValueError(
ValueError: Invalid coutour coord [ 1.  13.5] is generated, areas may contain boundaries or small holes
```

Could you please run the entire bundle (data preparation, training stage 0 and 1, evaluation, and inference) on your machine (from scratch) to make sure that all working?

KumoLiu:
Hi @drbeh, since the post-processing is not robust, I have encountered this problem before, if you draw the wrong data, you will find that it may contain some holes in some instances, which cannot be avoided in the whole post-processing. Thanks!

KumoLiu:
1. I searched further for the cause of the error and found that the default dtype setting in `Watershed` and `GenerateWatershedMarkers` is not very reasonable and causes problems when the number of instances exceeds 256. I create a PR to change the default value. https://github.com/Project-MONAI/MONAI/pull/5732
2.  Since the dtype value changed, we may change `SaveImage` in Inference or remove it. May need @drbeh to confirm here. Also in the[ tutorial repo](https://github.com/Project-MONAI/tutorials/blob/main/pathology/hovernet/inference.py).
https://github.com/Project-MONAI/model-zoo/blob/b54e8b0b45fdad456e005c4100548dfbcc162cf8/models/pathology_nuclei_segmentation_classification/configs/inference.json#L105-L122
3. I also added a workaround way to prevent the issue when the instance contains holes.
https://github.com/Project-MONAI/MONAI/blob/d1cdf416f3041c09cda4cd17afaa6ba97318ef0f/monai/apps/pathology/transforms/post/array.py#L735
I also take a look at the wrong case, which instance looks like this.
Screen Shot 2022-12-14 at 14 27 00

models/pathology_nuclei_segmentation_classification/large_files.yml
```diff
@@ -1,5 +1,5 @@
 large_files:
   - path: "models/model.pt"
-    url: "https://drive.google.com/uc?id=16CLRwqwxsToEmH-rSUOn2CSj_KmByGbp"
-    hash_val: "c00b643a4b4f9bdeb41ddf5a25cbea0a"
+    url: "https://drive.google.com/file/d/1Pznw-5AeoybEdjCrjmN-TzMHjWW0UjXn/view?usp=share_link"
```
bhashemian:
```suggestion
    url: "https://drive.google.com/uc?id=1Pznw-5AeoybEdjCrjmN-TzMHjWW0UjXn"
```

bhashemian (commented):
Thanks @KumoLiu! The whole bundle works on my machine.
@SachidanandAlle @yiheng-wang-nv do you have any other comments?

models/pathology_nuclei_segmentation_classification/large_files.yml
SachidanandAlle: have you guys verified the contours that you are generating.. after running hovernet sliding window inferer for some ROI?

KumoLiu: Hi @SachidanandAlle, I have verified it in jupyter notebook. But I didn't compare it with the results from the original repo and just had there is some inconsistency. Have created an issue to solve it.
https://github.com/Project-MONAI/MONAI/issues/5750
Thanks!

KumoLiu: Hi @drbeh, may I know why need to change the URL? Seems we already have the same format with the URLs from other bundles.
https://github.com/Project-MONAI/model-zoo/blob/fd95444a2feac5f76d9350a748deba5ba63311a2/models/brats_mri_segmentation/large_files.yml#L3

KumoLiu: Hi @SachidanandAlle, the inconsistency issue has been solved in https://github.com/Project-MONAI/MONAI/pull/5751. Do you have any other comments on this PR? Thanks!

drbeh: This way of representing a google drive link does not allow `gdown` to download it and instead it downloads the html page. Using `uc?id=` instead of `file/d/` can solve this issue.

@SachidanandAlle, how do you download the model weights for bundle when you are invoking it for a MOANILabel use case?

Nic-Ma: Hi @SachidanandAlle ,

We updated the bundle with pretrained weights from internal ImageNet instead of the previous public ImageNet.
Could you please help download and have a try?

Thanks.

SachidanandAlle: Infer looks good.. let me see if I can finetune the model over this..

SachidanandAlle: basic train workflow is ok via monailabel but the final model.pt (best metric should be correctly placed for others to consume/refresh)

image

Nic-Ma: Hi @KumoLiu ,

I think you missed the `key_metric_filename` arg in the train config, please refer to:
https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_ct_segmentation/configs/train.json#L255

Thanks.

SachidanandAlle: you can have any number of stages.. but final model which is ready for others to consume should be placed back into the correct `models` folder.  even inference scripts loads from models/mode.pt

https://github.com/KumoLiu/model-zoo/blob/hovernet-bundle/models/pathology_nuclei_segmentation_classification/configs/inference.json#L130

SachidanandAlle: Previous comment for reference: https://github.com/Project-MONAI/model-zoo/pull/245#discussion_r1043742493

SachidanandAlle: only this one issue.. otherwise all ok from myside.. to release the bundle.. verified for multi gpu as well

KumoLiu: Hi @SachidanandAlle, since we don't have validation data, the second stage reads the last epoch of the train, and the reference also reads the last epoch of the train of the second stage.
https://github.com/Project-MONAI/model-zoo/blob/700d1bdd4eca349ab06bd2cc75f1dff8247af911/models/pathology_nuclei_segmentation_classification/configs/train.json#L330-L331
Thanks!

SachidanandAlle: So you mean to say you can not produce models/model.pt once the training is finished?

KumoLiu: Hi @SachidanandAlle, Sorry for the confusion in the bundle, actually HoVerNet pipeline uses the last epoch model of the train as the final output model, instead of the best validation model. So you don't need to load the best validation key metric model, just use the `model.pt`.
And when you train stage 1,  we will automatically read the model from the first stage's fold. Actually, the final model after two stages is `model/model.pt`.
https://github.com/Project-MONAI/model-zoo/blob/700d1bdd4eca349ab06bd2cc75f1dff8247af911/models/pathology_nuclei_segmentation_classification/configs/train.json#L345-L350

SachidanandAlle: 👍

models/pathology_nuclei_segmentation_classification/docs/README.md
```diff
@@ -0,0 +1,132 @@
...
+```
+python -m monai.bundle run evaluating --meta_file configs/metadata.json --config_file configs/inference.json --logging_file configs/logging.conf
+```
+
+# Disclaimer
+
+This is an example, not to be used for diagnostic purposes.
+
+# References
+
```
yiheng-wang-nv, line 131: Hi @KumoLiu , could you add a License section in the readme file? You can refer to any other bundles.

Nic-Ma (approved):

models/pathology_nuclei_segmentation_classification/docs/README.md
KumoLiu: Have added, thanks!

yiheng-wang-nv (approved):

SachidanandAlle:
how you fine tune the model?
you have models/model.pt (as part of the bundle)

For stage 0 what is the value for `PRETRAIN_MODEL_URL` ?
For fine-tune, I guess you also need to provide stage 0 output model.. otherwise it will be always from scratch..
unless you want to claim that we can use models/model.pt as pretrained model for stage 0 as well

KumoLiu:
> how you fine tune the model? you have models/model.pt (as part of the bundle)
> 
> For stage 0 what is the value for `PRETRAIN_MODEL_URL` ?

Hi @SachidanandAlle, for most of the hyperparameters we set them according to the values in the [original repo](https://github.com/vqdang/hover_net).
Since we used internal dataset weight as our pretrained weight in stage 0 which is non-public, so we didn't put the link here.

KumoLiu:
That makes sense, I will add it.

SachidanandAlle:
> That makes sense, I will add it.

and by default your checkpoint loader for stage 0 will load that.. unless user specifies `PRETRAIN_MODEL_URL`.. will that possible and makes sense?  this is when user wants to fine tune the model (means, train additional/new samples over existing bundle model that we are releasing)

KumoLiu:
> and by default your checkpoint loader for stage 0 will load that.. unless user specifies `PRETRAIN_MODEL_URL`.. will that possible and makes sense? this is when user wants to fine tune the model (means, train additional/new samples over existing bundle model that we are releasing)

Hi @SachidanandAlle, I have added the final weight for stage 0. Users can simply train stage 1 from that weight. And also if they want to train from stage 0, they could specify a MODEL URL. If they don't specify a MODEL URL, then it will only train the decoder without pretrained weight.

SachidanandAlle:
> > and by default your checkpoint loader for stage 0 will load that.. unless user specifies `PRETRAIN_MODEL_URL`.. will that possible and makes sense? this is when user wants to fine tune the model (means, train additional/new samples over existing bundle model that we are releasing)
> 
> Hi @SachidanandAlle, I have added the final weight for stage 0. Users can simply train stage 1 from that weight. And also if they want to train from stage 0, they could specify a MODEL URL. If they don't specify a MODEL URL, then it will only train the decoder without pretrained weight.

Just verified.  Logistically Looks good now.

SachidanandAlle (approved):

Nic-Ma:
Thanks for the testing. Seems no other comments so far, @yiheng-wang-nv please merge it ASAP.

Thanks.

yiheng-wang-nv:
The blossom ci test failed due to breaking changes of hovernet. We may need to wait for monai 1.1 release and then merge this bundle. Hi @KumoLiu @Nic-Ma @drbeh , what do you think?

wyli:
perhaps you can set it to `1.1.0rc2` for now? I think rc2 has all code changes it needs

PR #246, opened 2022-11-22
Update Pathology Bundle Weights

bhashemian:
### Description
Update the pathology bundle weights and metrics

### Status
**Ready**


Nic-Ma (approved):

models/pathology_tumor_detection/eval/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.npy
+*.txt
```
yiheng-wang-nv: Hi @drbeh , could this `eval` folder be removed? I think it is produced after running the code.

bhashemian: Hi @yiheng-wang-nv, actually it is not being produced and causing error at the middle of inference when it want to write down the first probability map result.

yiheng-wang-nv: I see. I just checked the source code of `ProbMapProducer`, it would be nice to enhance it in the next version on making the directory first (like the `SaveImaged` transform does, https://github.com/Project-MONAI/MONAI/blob/a8f67e4278d46b49a87488a41ba38431883e8907/monai/data/folder_layout.py#L95)

What do you think? cc @Nic-Ma @wyli

Nic-Ma: Hi @drbeh ,

I would suggest to remove the `eval` folder in the bundle to be consistent with others, if users face this issue when executing the inference, just let them create the target folder.
What do you think?

Thanks.

bhashemian: @yiheng-wang-nv I removed eval folder.

yiheng-wang-nv (approved):

yiheng-wang-nv:
Hi @drbeh  , could you please help to double check the authority of the new weights? The CI test failed due to download issue. A part of logs is like below:

 Cannot retrieve the public link of the file. You may need to change the permission to 'Anyone with the link', or have had many accesses.

RuntimeError: Download of file from https://drive.google.com/uc?id=1Y-kLanna3osAEqZHiLeh5nOkAKcSU0uc to models/pathology_tumor_detection/models/model.pt failed due to network issue or denied permission.

PR #267, opened 2022-12-02
add detection inferer

### LINKED ISSUE #268: Add detection inferer to make inference more straightforward
No detection inferer. Inference not straightforward and not easy for deploy

Add detection inferer to make inference more straightforward

**Additional context**

Can-Zhao:

Fixes #268 .

### Description
Add detection inferer and change bundle accordingly

### Status


Can-Zhao:
I put detection inferer here for now. After we finalize it, I will move it to MONAI core

models/lung_nodule_ct_detection/scripts/evaluator.py
```diff
@@ -117,12 +116,12 @@ def __init__(
...
         device: torch.device,
         val_data_loader: Iterable | DataLoader,
-        detector: torch.nn.Module,
+        network: RetinaNetDetector,
         epoch_length: int | None = None,
         non_blocking: bool = False,
         prepare_batch: Callable = detection_prepare_val_batch,
         iteration_update: Callable[[Engine, Any], Any] | None = None,
-        inferer: Inferer | None = None,
+        inferer: RetinaNetInferer | None = None,
```
yiheng-wang-nv, line 124: Hi @Can-Zhao ,  thanks for the updates. I'd like to ask a question: with this change, can we get exactly the same inference/validate output?  I suspect the results are different, thus should the performance (on val set) be updated in readme too? If the score decreased, may also need to post the comparisons to let users know what are the differences, thus they are able to select which version to use. What do you think?

models/lung_nodule_ct_detection/scripts/detection_inferer.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        self.detector.network = network
+        self.detector.training = self.detector.network.training
+
+        if self.sliding_window_size is not None:
+            # if image smaller than sliding window roi size, no need to use sliding window inferer
+            # use sliding window inferer only when image is large
+            use_inferer = not all([data_i[0, ...].numel() < self.sliding_window_size for data_i in inputs])
+        else:
+            use_inferer = False
```
wyli:
```suggestion
        # if image smaller than sliding window roi size, no need to use sliding window inferer
        # use sliding window inferer only when image is large
        use_inferer = self.sliding_window_size is not None and not all([data_i[0, ...].numel() < self.sliding_window_size for data_i in inputs])
```

wyli (approved):

models/lung_nodule_ct_detection/scripts/evaluator.py
Can-Zhao: Thank you! Very good point. I run evaluation with the new version, and the result is exactly same with the previous version.

models/lung_nodule_ct_detection/scripts/detection_inferer.py
Can-Zhao: Thank you! commited

Can-Zhao:
@wyli Do you think we should put detection_inferer.py and warmup_scheduler.py here or put it to MONAI core?

models/lung_nodule_ct_detection/scripts/evaluator.py
yiheng-wang-nv: Thanks @Can-Zhao . Following up with this question, I also wish to ask about the required GPU memory. Is there any change of it? I'm not familiar with the algorithm, but I guess input the whole 3d image rather than a patch into the inferer may take more GPU memory. Is it still enough to use a 16GB GPU (mentioned in the readme)?
I'm fine with other places, and will approve this PR after resolve this question, thanks!

Can-Zhao: Thank you Yiheng. The GPU memory is the same. This script is just a wrapper. There is no fundamental change here. It did not input the whole 3d image. The GPU still takes a patch

yiheng-wang-nv: I see. Sorry I misunderstood the change. I will leave a comment in the place I feel confused.

models/lung_nodule_ct_detection/configs/inference.json
```diff
@@ -135,14 +135,8 @@
         "collate_fn": "$monai.data.utils.no_collation"
     },
     "inferer": {
-        "_target_": "SlidingWindowInferer",
-        "roi_size": [
-            240,
```
yiheng-wang-nv, line 140: Hi @Can-Zhao , I cannot understand this line. With the change, how to adjust the roi size of the inferer?
According to this line (https://github.com/Project-MONAI/model-zoo/blob/6a893936088192810f6f8d8f30e5fce2e8faef4a/models/lung_nodule_ct_detection/scripts/detection_inferer.py#L38), it depends on the `detector`, but it has the patch size `512, 512, 192` (https://github.com/Project-MONAI/model-zoo/pull/267/files#diff-10c2358610640d3b97968476b3699b54d3433c6d17689a196a8a6a3fcb844e39R16)

Can-Zhao: Thanks! In this case, 512, 512, 192 will be the self.detector.inferer.roi_size

yiheng-wang-nv (approved):

PR #277, opened 2022-12-15
Update GDrive URLs

bhashemian:
### Description
This PR makes the GDrive URL of the model weights (in `large_files.yml`)to be downloadable using [`gdown`](https://github.com/wkentaro/gdown), which requires using `/uc?id=` instead of `/file/d/`.

### Status


yiheng-wang-nv:
Hi @drbeh , I think `gdown` works fine with current links. Maybe you can check about the `--fuzzy` feature, we already use `gdown` in our ci test to check if a PR's download file works.

bhashemian:
> Hi @drbeh , I think `gdown` works fine with current links. Maybe you can check about the `--fuzzy` feature, we already use `gdown` in our ci test to check if a PR's download file works.

Sure, that would work too. thanks.

PR #313, opened 2023-02-13
148 Add name tag for bundles and related checks

### LINKED ISSUE #148: Display name for Models in metadata JSON
A display or "pretty" name for each Model in the metadata.json and adding a requirement that all future models also need a display name.

Add display names to each of the current MONAI Models metadata JSON so that we have the preferred name being displayed in the Model Zoo.

Creating a regex/replace function that replaces keywords like MRI, CT, NIST, UNETR, etc. within the current bundle names.  However, this would require customization as the Model Zoo continued to expand.

**Additional context**
Right now, when generating the Model Zoo page, the names being used are the bundle name with the first letter capitalized.  It would be better to have a proper display name that is a concise name used when generating the page.  That way, we have proper formatting for things like MRI or UNETR.

Example:
image

yiheng-wang-nv:
Fixes #148  .

### Description
This PR adds name tag for all bundles, as well as the related checks. A few doc errors are fixed when I did this work.

### Status
**Ready**

- [x] In-line docstrings updated.

### DIFF
--- a/models/pathology_nuclick_annotation/configs/metadata.json
+++ b/models/pathology_nuclick_annotation/configs/metadata.json
@@ -1,7 +1,8 @@
 {
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
-    "version": "0.0.4",
+    "version": "0.0.5",
     "changelog": {
+        "0.0.5": "add name tag",
         "0.0.4": "Fix evaluation",
         "0.0.3": "Update to use MONAI 1.1.0",
         "0.0.2": "Update The Torch Vision Transform",
@@ -14,8 +15,9 @@
         "nibabel": "4.0.1",
         "pytorch-ignite": "0.4.9"
     },
-    "task": "Pathology Nuclick segmentation",
-    "description": "A pre-trained model for Nuclei Classification within Haematoxylin & Eosin stained histology images",
+    "name": "Pathology nuclick annotation",
+    "task": "Pathology Nuclick annotation",
+    "description": "A pre-trained model for segmenting nuclei cells with user clicks/interactions",
     "authors": "MONAI team",
     "copyright": "Copyright (c) MONAI Consortium",
     "data_source": "consep_dataset.zip from https://warwick.ac.uk/fac/cross_fac/tia/data/hovernet",
--- a/models/valve_landmarks/configs/metadata.json
+++ b/models/valve_landmarks/configs/metadata.json
@@ -1,7 +1,8 @@
 {
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220729.json",
-    "version": "0.4.1",
+    "version": "0.4.2",
     "changelog": {
+        "0.4.2": "add name tag",
         "0.4.1": "modify dataset key name",
         "0.4.0": "update license files",
         "0.3.0": "Update to scripts",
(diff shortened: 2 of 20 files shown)


models/pathology_nuclick_annotation/configs/metadata.json
```diff
@@ -14,8 +15,9 @@
         "nibabel": "4.0.1",
         "pytorch-ignite": "0.4.9"
     },
-    "task": "Pathology Nuclick segmentation",
-    "description": "A pre-trained model for Nuclei Classification within Haematoxylin & Eosin stained histology images",
+    "name": "Pathology nuclick annotation",
+    "task": "Pathology Nuclick annotation",
+    "description": "A pre-trained model for segmenting nuclei cells with user clicks/interactions",
```
yiheng-wang-nv, line 20: Hi @SachidanandAlle , here I did a few changes of the description according to the readme file, and it seems the old version is from the classification bundle.
Could you please help to double confirm these three nuclei bundles? The "name" tag added here is to define the display name in the webpage (https://monai.io/model-zoo.html). Thanks in advance!

wyli (approved):

models/valve_landmarks/configs/metadata.json
```diff
@@ -12,6 +13,7 @@
     "pytorch_version": "1.13.0",
     "numpy_version": "1.21.2",
     "optional_packages_version": {},
+    "name": "Valve landmarks regression",
```
yiheng-wang-nv, line 16: Hi @ericspod , the added "name" tag is to represent the display name of the bundle in the webpage. Could you please help to check the names? Feel free to make any modifications of all of your bundles. Thanks in advance!

models/pathology_nuclick_annotation/configs/metadata.json
SachidanandAlle: LGTM

models/valve_landmarks/configs/metadata.json
ericspod, line 16:
```suggestion
    "name": "Valve Landmarks Regression",
```
I would suggest here and elsewhere that names should be in title case as this would look best on the model zoo website and elsewhere.

models/ventricular_short_axis_3label/configs/metadata.json
```diff
@@ -14,6 +15,7 @@
         "nibabel": "3.2.1",
         "pytorch-ignite": "0.4.8"
     },
+    "name": "Ventricular short axis 3 label segmentation",
```
ericspod, line 18:
```suggestion
    "name": "Ventricular Short Axis 3 Label Segmentation",
```

models/valve_landmarks/configs/metadata.json
ericspod: Hi @yiheng-wang-nv  I suggested using capital case for those I contributed but I didn't want to change other people's names.

Nic-Ma: Hi @ericspod @yiheng-wang-nv , I am OK with both naming formats, but please try to unify all the bundle names.

Thanks.

yiheng-wang-nv: Thanks @ericspod , as @Nic-Ma suggested, I'm not sure if we should capitalize all words, thus I followed the rule of our current webpage and only keep the first word capital. What do you think?

yiheng-wang-nv: Hi @ericspod , how about merging this PR, and if needing to use capital case on all words, feel free to submit a new ticket to discuss? Thanks!

ericspod: @yiheng-wang-nv  Yes let's look at that later and put this through for now.

PR #316, opened 2023-02-17
Update SwinUNETR Bundle Params to Support torch.jit.trace Conversion

tangy5:
Disable use_checkpoint param to support torchscript conversion by torch.jit.trace. 

Tested with the bundle, functions and performance are the same.

### DIFF
--- a/models/swin_unetr_btcv_segmentation/configs/inference.json
+++ b/models/swin_unetr_btcv_segmentation/configs/inference.json
@@ -15,7 +15,7 @@
         "in_channels": 1,
         "out_channels": 14,
         "feature_size": 48,
-        "use_checkpoint": true
+        "use_checkpoint": false
     },
     "network": "$@network_def.to(@device)",
     "preprocessing": {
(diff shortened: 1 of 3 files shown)


yiheng-wang-nv (approved):

binliunls:
Hi @tangy5 ,
     Parameters in `train.json` won't affect the trt convert, since we only do the inference with trt. So if the `use_checkpoint` parameter in the `train.json` is good for training. Maybe we shoud not change it?
Thanks,
Bin

yiheng-wang-nv:
> Hi @tangy5 , Parameters in `train.json` won't affect the trt convert, since we only do the inference with trt. So if the `use_checkpoint` parameter in the `train.json` is good for training. Maybe we shoud not change it? Thanks, Bin

Hi @binliunls , thanks for the comments. Hi @tangy5 , this PR is merged, and if need to modify the `train.json` back as Bin suggested, feel free to submit a new PR, thanks in advance!

Issue #338, opened 2023-03-28
Show figures via using url rather than add files


yiheng-wang-nv:
The following bundles need to avoid adding files to display figures in readme:

1. `pathology_nuclick_annotation` @tangy5 
2. `pathology_nuclei_classification` @KumoLiu 
3. `wholeBody_ct_segmentation` @tangy5

PR #341, opened 2023-03-31
Update `pathology_nuclei_classification` benchmark

KumoLiu:
Part of #338 .

### Description

- [x]  Retrain pathology_nuclei_segmentation_classification on A100.

- [x]  Update figures with URL and remove unused files.

### Status
**Work in progress**


KumoLiu:
Hi @Nic-Ma @SachidanandAlle, both experiments showed a performance drop, if you think it's acceptable, I will update the model weights ASAP.
Thanks in advance!

SachidanandAlle (approved):
The drop or variation is mostly due to random selection of test vs train samples.  And also the samples for some of cell types (Other and Inflammatory) is too low.

bhashemian:
@KumoLiu thanks for re-running and updating the results. Is this the result of one experiment?

KumoLiu:
> @KumoLiu thanks for re-running and updating the results. Is this the result of one experiment?

Hi @drbeh, what did you mean by "one experiment"? Only trained with one seed?

bhashemian:
> > @KumoLiu thanks for re-running and updating the results. Is this the result of one experiment?
> 
> Hi @drbeh, what did you mean by "one experiment"? Only trained with one seed?

Yes @KumoLiu, I meant if the experiment is based on one seed.

KumoLiu:
> > > @KumoLiu thanks for re-running and updating the results. Is this the result of one experiment?
> > 
> > 
> > Hi @drbeh, what did you mean by "one experiment"? Only trained with one seed?
> 
> Yes @KumoLiu, I meant if the experiment is based on one seed.

Yes.

yiheng-wang-nv:
Hi @KumoLiu , as several bundles are updated with new workflow, may need to resolve the conflicts in this PR, thanks!

models/pathology_nuclei_segmentation_classification/docs/README.md
```diff
@@ -6,9 +6,7 @@ A pre-trained model for simultaneous segmentation and classification of nuclei w
 
 The model is trained to simultaneous segment and classify nuclei. Training is done via a two-stage approach. First initialized the model with pre-trained weights on the [ImageNet dataset](https://ieeexplore.ieee.org/document/5206848), trained only the decoders for the first 50 epochs, and then fine-tuned all layers for another 50 epochs. There are two training modes in total. If "original" mode is specified, it uses [270, 270] and [80, 80] for `patch_size` and `out_size` respectively. If "fast" mode is specified, it uses [256, 256] and [164, 164] for `patch_size` and `out_size` respectively. The results we show below are based on the "fast" model.
```
yiheng-wang-nv: Should the source of pretrained weights be modified here?

models/pathology_nuclei_classification/configs/metadata.json
```diff
@@ -2,6 +2,7 @@
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
     "version": "0.0.6",
```
yiheng-wang-nv: need to use 0.0.7

yiheng-wang-nv (approved):

yiheng-wang-nv:
> 

I tried to run the training config multiple times, and confirmed that this bundle is non-determinism.

bhashemian:
> I tried to run the training config multiple times, and confirmed that this bundle is non-determinism.

@yiheng-wang-nv the `benchmark` in  `torch.backends.cudnn` is set to `True`, which makes any convolution-based pipeline non-deterministic:
https://github.com/Project-MONAI/model-zoo/blob/ec15cddd5998c935b7106f29520b3f66a29b16fe/models/pathology_nuclei_classification/configs/train.json#L347

In order to make sure that we get the same result, we should set it to `False`. I see the same thing in other bundles too.

yiheng-wang-nv:
> > I tried to run the training config multiple times, and confirmed that this bundle is non-determinism.
> 
> @yiheng-wang-nv the `benchmark` in `torch.backends.cudnn` is set to `True`, which makes any convolution-based pipeline non-deterministic:
> 
> https://github.com/Project-MONAI/model-zoo/blob/ec15cddd5998c935b7106f29520b3f66a29b16fe/models/pathology_nuclei_classification/configs/train.json#L347
> 
> In order to make sure that we get the same result, we should set it to `False`. I see the same thing in other bundles too.

Hi @drbeh , thanks and I checked and confirmed that when using single GPU and set `torch.backends.cudnn.benchmark = False` can get deterministic result, but training on multiple GPUs cannot ensure that.

In addition, this command may not enough on some pipelines because of using pytorch unsupported layers/algorithms ..., users can set [torch.use_deterministic_algorithms()](https://pytorch.org/docs/stable/generated/torch.use_deterministic_algorithms.html#torch.use_deterministic_algorithms) to ensure deterministic training (if having unsupported parts, will raise an error). MONAI also enables it (see: https://github.com/Project-MONAI/MONAI/blob/6a7f35b7e271cdf3b2c763ae35acf4ab25d85d23/monai/utils/misc.py#L316)

yiheng-wang-nv:
Hi @drbeh BTW,
> we should set it to False

I think it depends on each bundle's author, since set it to False may also reduce the performance according to the official doc: https://pytorch.org/docs/stable/notes/randomness.html#cuda-convolution-benchmarking  Authors may need to do some experiments and make trade off. In model zoo side, there is no hard requirement of which flag to set.

yiheng-wang-nv:
> Hi @drbeh BTW,
> 
> > we should set it to False
> 
> I think it depends on each bundle's author, since set it to False may also reduce the performance according to the official doc: https://pytorch.org/docs/stable/notes/randomness.html#cuda-convolution-benchmarking Authors may need to do some experiments and make trade off. In model zoo side, there is no hard requirement of which flag to set.

Update: we aligned with the determinism issue offline, and will address it in a separate PR and ticket. This PR will be merged first.

PR #363, opened 2023-04-17
362 Update deterministic settings on several bundles

yiheng-wang-nv:
Fixes #362 .

### Description
This PR adds deterministic settings on several bundles mentioned in #362 , and for unsupported bundles, I added some notes on readme files.

### Status
**Ready**

### Please ensure all the checkboxes:

- [x] Codeformat tests passed locally by running `./runtests.sh --codeformat`.
- [x] In-line docstrings updated.
- [x] Update `version` and `changelog` in `metadata.json` if changing an existing bundle.
- [x] Please ensure the naming rules in config files meet our requirements (please refer to: `CONTRIBUTING.md`).
- [x] Ensure versions of packages such as `monai`, `pytorch` and `numpy` are correct in `metadata.json`.
- [x] Descriptions should be consistent with the content, such as `eval_metrics` of the provided weights and TorchScript modules.
- [x] Files larger than 25MB are excluded and replaced by providing download links in `large_file.yml`.
- [x] Avoid using path that contains personal information within config files (such as use `/home/your_name/` for `"bundle_root"`).

PR #364, opened 2023-04-17
Update the endoscopic tool segmentation weight

binliunls:
### Description
Update the endoscopic tool segmentation weight.

### Status


models/endoscopic_tool_segmentation/large_files.yml
```diff
@@ -1,9 +1,9 @@
 large_files:
   - path: "models/model.pt"
-    url: "https://drive.google.com/file/d/19yS3t2oLBiB7wT-qeQ82da95VJs_vzRK/view?usp=share_link"
-    hash_val: "9d568aaf7188b7b3f254f87c8656d7fe"
+    url: "https://drive.google.com/file/d/1_VbLCkQBbcHC1I38aDN4nkomVOJYpGCW/view?usp=sharing"
```
yiheng-wang-nv: Hi @binliunls , thanks for the changes. Are the weights here reproducible? If not, may still need to retrain it after #363 is merged.

models/endoscopic_tool_segmentation/docs/README.md
```diff
@@ -3,6 +3,12 @@ A pre-trained model for the endoscopic tool segmentation task and is trained usi
 
 The [PyTorch model](https://drive.google.com/file/d/19yS3t2oLBiB7wT-qeQ82da95VJs_vzRK/view?usp=share_link) and [torchscript model](https://drive.google.com/file/d/1cDZ3Jr7mhpzdzaFyz8yHNowH8k0T1VZz/view?usp=share_link) are shared in google drive. Details can be found in large_files.yml file. Modify the "bundle_root" parameter specified in configs/train.json and configs/inference.json to reflect where models are downloaded. Expected directory path to place downloaded models is "models/" under "bundle_root".
 
+Either an [ImageNet](https://ieeexplore.ieee.org/document/5206848) pre-trained or a local pre-trained encoder weight could be loaded to the model for training, which is controlled by the `use_imagenet_pretrain` parameter in the `configs/train.json` file. If using local pre-trained weight, please update the `load_path`  parameter of the `CheckpointLoader` in the `configs/train.json` with the local weight path.
```
yiheng-wang-nv: Usually say pretrained weights (not weight)

yiheng-wang-nv: I think this part should be rephrased to be more clear. Same issue for the `pathology_nuclei_segmentation_classification` bundle. I tried to modify that bundle as you can see in: https://github.com/Project-MONAI/model-zoo/pull/363/files#diff-a739ca80d65b7994071d4c6f0fc9a2397f698b958b98f1d314056f1fe260d2acR5
but it may still have some issues and may need some experts to double confirm. cc @KumoLiu

models/endoscopic_tool_segmentation/large_files.yml
binliunls: Hi @yiheng-wang-nv ,
    Thank you for the info. This is just the first drop with private dataset. I will keep updating this bundle until it achieve the previous IoU and will update it after the #363 merged.

Thanks,
Bin

models/endoscopic_tool_segmentation/docs/README.md
yiheng-wang-nv: Update: we've aligned and @zephyrie will help to review it (may submit a separate PR) thus we don't need to worry too much about it on this PR

models/endoscopic_tool_segmentation/docs/README.md
```diff
@@ -1,10 +1,17 @@
...
 
 ![image](https://developer.download.nvidia.com/assets/Clara/Images/monai_endoscopic_tool_segmentation_workflow.png)
 
+## Pre-trained weights
+A pre-trained encoder weights would benefit the training accuracy. Either the [ImageNet](https://ieeexplore.ieee.org/document/5206848) pre-trained or a local pre-trained encoder weights could be loaded to the model for the segmentation training, which is controlled by the `use_imagenet_pretrain` parameter in the `configs/train.json` file. If using the local pre-trained weights, please update the `load_path`  parameter of the `CheckpointLoader` in the `configs/train.json` file with the local weights path.
+
+- We trained the encoder with pre-trained weights from some internal data.
+
+- The ImageNet pre-trained encoder repo has pre-trained weights which is for non-commercial use. Each user is responsible for checking the content of models/datasets and the applicable licenses and determining if suitable for the intended use. The license for the pre-trained model is different than MONAI license. Please check the source where these weights are obtained from: <https://github.com/lukemelas/EfficientNet-PyTorch>
```
yiheng-wang-nv: https://github.com/Project-MONAI/model-zoo/tree/dev/models/pathology_nuclei_segmentation_classification could you take this bundle's readme for reference to rephrase this section?

models/endoscopic_tool_segmentation/docs/README.md
```diff
@@ -1,10 +1,17 @@
...
 
+## Pre-trained weights
+A pre-trained encoder weights would benefit the training accuracy. Either the [ImageNet](https://ieeexplore.ieee.org/document/5206848) pre-trained or a local pre-trained encoder weights could be loaded to the model for the segmentation training. Whether to use the ImageNet pre-trained weights is controlled by the `use_imagenet_pretrain` parameter in the `configs/train.json` file. If using the local pre-trained weights, please update the `load_path`  parameter of the `CheckpointLoader` in the `configs/train.json` file with the local weights path.
+
+- We trained a pre-trained encoder weights using some internal data.
+
+- The [EfficientNet-PyTorch repo](https://github.com/lukemelas/EfficientNet-PyTorch) provides the ImageNet pre-trained encoder weights but for non-commercial use. The ImageNet pre-trained encoder weights would be loaded for training, if setting the `use_imagenet_pretrain` parameter to `True`. Each user is responsible for checking the content of the models/datasets and the applicable licenses and determining if suitable for the intended use.
```
yiheng-wang-nv: duplicate information between this line and https://github.com/Project-MONAI/model-zoo/pull/364/files#diff-1d6658e5d0f1616da1655e5be3b8a383c0c6a7e1e4782fd6210850c48444f00aR9
I will update the readme to address it

yiheng-wang-nv (approved):
I modified the readme file, and if there are any mis-understanding places, feel free to correct me via submitting a new PR, thanks @binliunls

models/endoscopic_tool_segmentation/configs/train.json
```diff
@@ -136,6 +137,16 @@
             "_target_": "SimpleInferer"
         },
         "handlers": [
+            {
+                "_target_": "CheckpointLoader",
+                "_disabled_": "@use_imagenet_pretrain",
+                "load_path": "/path/to/local/weight/model.pt",
```
SachidanandAlle, line 143: Can you please define the variable above and use.. instead of doing like this... with this difficult to override the load_path from command line..

@tangy5 we need this fix so that monailabel can pass override values for `load_path`

SachidanandAlle: otherwise you can mark use_imagenet_pretrain disabled by default.. and let user chose when to enable and provide the correct path

binliunls: Hi @SachidanandAlle ,
    For the second suggestion, did you mean to set the default value of `use_imagenet_pretrain` to `true`, so that the bundle could run with the imagenet pretrain weight by default? Hi @Nic-Ma , I think it would be easy to set the `use_imagenet_pretrain` parameter to `true` by default. What do you think about this?

Thanks,
Bin

Nic-Ma: Add @KumoLiu

KumoLiu: Hi @binliunls, I don't think we could set the default value of `use_imagenet_pretrain` to `true` here in the model zoo since there may be some issues with the terms of use. 
Thanks!

yiheng-wang-nv: I suggest to define a variable like "pretrained_weights_path", it is easier to override. As for the `use_imagenet_pretrain`, I think we should make it `False` to avoid license related issues.

Hi @binliunls , another suggest is that with default settings, users must provide an existing weights, otherwise error will happen in the handler.

I think the tumor detection bundle is a good example: https://github.com/Project-MONAI/model-zoo/tree/21d85dabfb38d013d15955ad9b1556f25bbdde3c/models/pathology_tumor_detection#pretrained-weights

Maybe we can remove the handler from the config file, to ensure that if not using any pretrained weights, users can still run the bundle. What do you think?

binliunls: Hi @yiheng-wang-nv and @KumoLiu ,
     I am a little bit confused about your advices?  Now in the endoscopic tool segmentation bundle, if `use_imagenet_pretrain` is set to `true`, the bundle could run successfully with the imagenet pretrain weight, which may have some license issue. If `use_imagenet_pretrain` is set to `false`, then a pretrained weight must be provided to run the bundle. Here are two different kind of modifications I think I can do:
1, Directly update the default value of `use_image_pretrain` to `true`, so that the bundle could run directly after downloaded.
2, Keep the `use_image_pretrain` to `false`, remove the `CheckpointLoader` handler and add an explanation about how to load pretrained model with the `CheckpointLoader`  in the READ file.

I personally think the second one is better. What's your opinions?

Thanks,
Bin

yiheng-wang-nv: Hi @binliunls , I think your 2nd method is exactly what I mean : )

Nic-Ma: I agree with Yiheng.

Thanks.

Nic-Ma: Please be consistent with the HoVerNet bundle. CC @KumoLiu

PR #369, opened 2023-04-18
fix finalize issue

yiheng-wang-nv:
Fixes # .

### Description
This PR fixes finalize command issue in multi-gpu config files.

### Status

### DIFF
--- a/models/endoscopic_inbody_classification/configs/multi_gpu_train.json
+++ b/models/endoscopic_inbody_classification/configs/multi_gpu_train.json
@@ -34,6 +34,6 @@
         "$@train#trainer.run()"
     ],
     "finalize": [
-        "$dist.destroy_process_group()"
+        "$dist.is_initialized() and dist.destroy_process_group()"
     ]
 }
--- a/models/brats_mri_segmentation/configs/metadata.json
+++ b/models/brats_mri_segmentation/configs/metadata.json
@@ -1,7 +1,8 @@
 {
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
-    "version": "0.4.1",
+    "version": "0.4.2",
     "changelog": {
+        "0.4.2": "fix mgpu finalize issue",
         "0.4.1": "add non-deterministic note",
         "0.4.0": "adapt to BundleWorkflow interface",
         "0.3.9": "black autofix format and add name tag",
--- a/models/brats_mri_segmentation/configs/multi_gpu_train.json
+++ b/models/brats_mri_segmentation/configs/multi_gpu_train.json
@@ -35,6 +35,6 @@
         "$@train#trainer.run()"
     ],
     "finalize": [
-        "$dist.destroy_process_group()"
+        "$dist.is_initialized() and dist.destroy_process_group()"
     ]
 }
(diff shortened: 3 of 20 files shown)


KumoLiu (approved):

models/endoscopic_inbody_classification/configs/multi_gpu_train.json
```diff
@@ -34,6 +34,6 @@
         "$@train#trainer.run()"
     ],
     "finalize": [
-        "$dist.destroy_process_group()"
+        "$dist.is_initialized() and dist.destroy_process_group()"
```
wyli, line 37: Do we need 'dist.is_available() and dist.is_initialized()'?

yiheng-wang-nv: Hi @wyli , thanks for the suggestion. It seems `dist.is_available()` is always `True` if the distributed package is avaiable. I'm not sure if we need to add it. Correct me if I misunderstand anything here, thanks!

wyli: if we don't check it here, I think we'll have issue like this one at some point https://github.com/facebookresearch/maskrcnn-benchmark/issues/280

Nic-Ma: Hi @wyli ,

I think this PR is mainly to avoid duplicated "destroy dist" operations if launching several train workflow and evaluate workflow together.
If the user's program can run until this "finalize" section, at least the `dist` should be OK? So maybe no need to add the "dist.is_available()".

Thanks.

Nic-Ma (approved):

models/endoscopic_inbody_classification/configs/multi_gpu_train.json
wyli: from the python API, now there's the possibility that `finalize` getting called at any time, correct?

wyli: it might be an issue for the initialize as well

yiheng-wang-nv: Hi @wyli ,
> there's the possibility that finalize getting called at any time

that's true and if not having the distributed package, use `dist.is_available() and dist.is_initialized()` will not raise any error.
However, in this case I think the error is informative and useful (can let users know ddp is not suitable). What do you think?

wyli: The error will be AttributeError: module 'torch.distributed' has no attribute 'is_initialized'

i don't think its informative.

Nic-Ma: Hi @wyli ,

I understand your point.
If using `dist.is_available() and dist.is_initialized()`, it will not raise any error, how to raise a proper error?
We didn't find a better expression, do you have any suggestion for this expression?
CC @holgerroth .

Thanks in advance.

wyli: I think it's fine to not raising an error in `finalize`. you can consider a better error message for the `initialize` section. maybe it's out of scope of this PR though.

wyli: ok sure, we spend too much time on this. please feel free to ignore my comments and merge this.

PR #377, opened 2023-04-20
Deterministic retrain 3 bundles

tangy5:


models/swin_unetr_btcv_segmentation/docs/README.md
```diff
@@ -55,11 +55,11 @@ The training as performed with the following:
 Dice score was used for evaluating the performance of the model. This model achieves a mean dice score of 0.8120
```
yiheng-wang-nv: plz double confirm if the value is still the same

yiheng-wang-nv: same for metadata  https://github.com/Project-MONAI/model-zoo/pull/377/files#diff-e196d6297358b2c9bf73633c09bc6dc1bc38d4b688a0bb1a52457d4c6e0a572fL45

yiheng-wang-nv (approved):

PR #386, opened 2023-04-24
Update detection bundle with TensorRT

### LINKED ISSUE #385: Support the lung_nodule_ct_segmentation bundle with TensorRT
The lung_nodule_ct_segmentation bundle doesn't support the TensorRT inference for now, since there are some pythonic logics in this bundle. It would be better if the bundle could be accelerated by the TensorRT conversion. The blocking issues are:
1. The `detector` takes a torch `network` as input, which has some necessary attributes like `num_anchors`, `num_class`, `spatial_dim` and so on. But the model exported by TensorRT doesn't retain these attributes. 
2. The `detector` have two ways to do inference. If the input size is larger than the presetting `infer_patch_size`, a `slide_window_inferer` is chosen for the inference. Otherwise, the input is directly fed to the network for inference. This causes the inputs of the network have different spatial sizes,   `infer_patch_size` or the input size (if the input size is smaller than the  `infer_patch_size`). However the TensorRT model only takes fixed size inputs (except for the batch size).

1. Update the `detector` with input parameters to replace the `network` attributes.
2. Add a parameter to contoll whether force the inference with `slide_window_inferer`.

**Additional context**

binliunls:
Fixes #385  .

### Description
Update the detection bundle with ONNX-TensorRT conversion.

### Status
**Work in progress**


models/lung_nodule_ct_detection/configs/metadata.json
```diff
@@ -1,7 +1,8 @@
 {
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
-    "version": "0.5.5",
+    "version": "0.5.6",
     "changelog": {
+        "0.5.6": "add the ONNX-TensorRT way of model conversion",
```
yiheng-wang-nv, line 5: could you add  this bundle into:
https://github.com/Project-MONAI/model-zoo/blob/dev/ci/bundle_custom_data.py#L39

models/lung_nodule_ct_detection/scripts/detection_inferer.py
```diff
@@ -25,14 +25,19 @@ class RetinaNetInferer(Inferer):
     Args:
         detector: the RetinaNetDetector that converts network output BxCxMxN or BxCxMxNxP
             map into boxes and classification scores.
+        force_inferer: whether to force using a SlidingWindowInferer to do the inference.
```
yiheng-wang-nv: Hi @Can-Zhao , may need your help to review this file. I'm not sure if the naming here is suitable. Thanks in advance.

models/lung_nodule_ct_detection/docs/README.md
```diff
@@ -63,6 +63,33 @@ The validation accuracy in this curve is the mean of mAP, mAR, AP(IoU=0.1), and
 
 ![A graph showing the detection val accuracy](https://developer.download.nvidia.com/assets/Clara/Images/monai_retinanet_detection_val_acc_v2.png)
 
+#### TensorRT speedup
+The `lung_nodule_ct_detection` bundle supports the TensorRT acceleration through the ONNX-TensorRT way. The table below shows the speedup ratios benchmarked on an A100 80G GPU. Please note the `force_inferer` in the `inference.json` must be set to `true` when doing inference with the TensorRT model. This forces the bundle using the `SlidingWindowInferer` during inference and keeps the input spatial size of the network invariable. Otherwise, given an input with spatial size less than the `infer_patch_size`, the input spatial size of the network would be changed.
```
yiheng-wang-nv: Otherwise, if given an input ...

models/lung_nodule_ct_detection/scripts/detection_inferer.py
Can-Zhao: Looks good to me! Thank you!

Can-Zhao:
@binliunls Do we need to retrain in this case? Or would the existing trained weights work with TensorRT inference?

Can-Zhao (approved):

models/lung_nodule_ct_detection/scripts/detection_inferer.py
Nic-Ma: I feel this arg name is very confusing, would suggest to others, for example: `force_sliding_window`.
@binliunls @Can-Zhao What do you think?

Thanks.

models/lung_nodule_ct_detection/configs/inference.json
```diff
@@ -13,6 +13,14 @@
     "test_datalist": "$monai.data.load_decathlon_datalist(@data_list_file_path, is_segmentation=True, data_list_key='validation', base_dir=@dataset_dir)",
     "device": "$torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')",
     "amp": true,
+    "spatial_dims": 3,
+    "num_classes": 1,
+    "force_inferer": true,
```
Nic-Ma: May I know why you put these 3 parameters as global vars here?

Thanks.

models/lung_nodule_ct_detection/scripts/detection_inferer.py
Can-Zhao: Yes, force_sliding_window sounds better

models/lung_nodule_ct_detection/configs/inference.json
binliunls: Hi @Nic-Ma ,
     Because the `spatial_dims` and `num_classes` parameters are used in both `network_def` and `detector`, I think it would be better to use a parameter to control them instead of directly specifying the value in the `network_def` and `detector`.
     For the `force_inferer` (or `force_sliding_window`) parameter, it is set as a global vars so that users can easily use it in the command line, like `python -m monai.bundle run ... --force_sliding_window "True"`

Thanks,
Bin

binliunls:
> @binliunls Do we need to retrain in this case? Or would the existing trained weights work with TensorRT inference?

Hi Can, no need to retrain the model, the TensorRT model only exports the trained model to a more efficient one and does the inference with it.

models/lung_nodule_ct_detection/docs/README.md
```diff
@@ -63,6 +63,33 @@ The validation accuracy in this curve is the mean of mAP, mAR, AP(IoU=0.1), and
 
 ![A graph showing the detection val accuracy](https://developer.download.nvidia.com/assets/Clara/Images/monai_retinanet_detection_val_acc_v2.png)
 
+#### TensorRT speedup
```
yiheng-wang-nv, line 66: Hi @zephyrie , could you please help to review this section? Thanks in advance!

binliunls:
Hi @Can-Zhao ,
       Could you please take a look at [this line](https://github.com/Project-MONAI/model-zoo/blob/ef6f861b2c14add848ba45d55dd1bc7c8a31ea78/models/lung_nodule_ct_detection/scripts/detection_inferer.py#L32). The `*args, **kwargs` part seems useless. I would like to update it from `def __init__(self, detector: RetinaNetDetector, *args, **kwargs) -> None` to `def __init__(self, detector: RetinaNetDetector, force_sliding_window: bool=False) -> None`. What do you think about it.

Thanks,
Bin

Can-Zhao:
> Hi @Can-Zhao , Could you please take a look at [this line](https://github.com/Project-MONAI/model-zoo/blob/ef6f861b2c14add848ba45d55dd1bc7c8a31ea78/models/lung_nodule_ct_detection/scripts/detection_inferer.py#L32). The `*args, **kwargs` part seems useless. I would like to update it from `def __init__(self, detector: RetinaNetDetector, *args, **kwargs) -> None` to `def __init__(self, detector: RetinaNetDetector, force_sliding_window: bool=False) -> None`. What do you think about it.
> 
> Thanks, Bin

Yes, it is useless.

yiheng-wang-nv (approved):
LGTM, after @zephyrie 's double confirm we can merge this PR

yiheng-wang-nv:
Update: @zephyrie reviewed the readme, and sent me some suggestions offline, I will push a new commit to this PR to address that.
For other bundles that have similar descriptions, I will modify them as well within this PR.

models/pathology_tumor_detection/configs/metadata.json
```diff
@@ -1,7 +1,8 @@
 {
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
-    "version": "0.5.1",
+    "version": "0.5.2",
     "changelog": {
+        "0.5.2": "update TensorRT descriptions",
```
binliunls, line 5: Should the monai version of this bundle be changed to `1.2.0rc5`? Same question for other bundles that don't change to rc5.
Thanks,
Bin

binliunls:
LGTM, thanks for the update.

PR #394, opened 2023-04-26
Update TensorRT part of spleen ct segmentation bundle

binliunls:
### Description
Update the TensorRT part of the spleen_ct_segmentation bundle to ONNX-TensorRT only.

### Status
**Work in progress**

### DIFF
--- a/ci/bundle_custom_data.py
+++ b/ci/bundle_custom_data.py
@@ -40,4 +40,5 @@ include_verify_onnx_tensorrt_list = [
     "brats_mri_segmentation",
     "endoscopic_inbody_classification",
     "spleen_deepedit_annotation",
+    "spleen_ct_segmentation",
 ]
--- a/models/spleen_ct_segmentation/docs/README.md
+++ b/models/spleen_ct_segmentation/docs/README.md
@@ -45,13 +45,12 @@ Dice score is used for evaluating the performance of the model. This model achie
 ![A graph showing the validation mean Dice over 1260 epochs.](https://developer.download.nvidia.com/assets/Clara/Images/monai_spleen_ct_segmentation_val.png)
 
 #### TensorRT speedup
-The `spleen_ct_segmentation` bundle supports the TensorRT acceleration. The table below shows the speedup ratios benchmarked on an A100 80G GPU.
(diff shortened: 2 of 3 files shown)


ci/bundle_custom_data.py
```diff
@@ -40,4 +40,5 @@
     "brats_mri_segmentation",
     "endoscopic_inbody_classification",
     "spleen_deepedit_annotation",
+    "spleen_ct_segmentation",
```
yiheng-wang-nv, line 43: should this list be updated?

yiheng-wang-nv: https://github.com/Project-MONAI/model-zoo/pull/394/files#diff-74b06043d5901ecb806cd7a85f3bd03a89b63c0e85ec3ca125c39ae4aada319fL35

binliunls: Hi @yiheng-wang-nv ,
     Actually the spleen_ct_segmentation bundle supports both ONNX-TensorRT and Torch-TensorRT ways of conversion. However the Torch-TensorRT way cannot achieve a decent speedup ratio. Therefore, for the short term usage, the ONNX-TensorRT way is chosen to export the model. But from the long term point of view, it will back to the Torch-TensorRT way finally (as long as it achieves a normal speedup ratio). So I think we'd better keep both TensorRT checks, in case there are some regression issues of Torch-TensorRT. What do you think about this.

Thanks,
Bin

models/spleen_ct_segmentation/docs/README.md
```diff
@@ -61,13 +60,15 @@ Where:
 - `speedup amp`, `speedup fp32` and `speedup fp16` are the speedup ratios of corresponding models versus the PyTorch float32 model
 - `amp vs fp16` is the speedup ratio between the PyTorch amp model and the TensorRT float16 based model.
 
+Currently, the only available method to accelerate this model is through ONNX-TensorRT. However, the Torch-TensorRT method is under development and will be available in the near future.
```
yiheng-wang-nv, line 63: Hi @binliunls  plz also update https://github.com/Project-MONAI/model-zoo/pull/394/files#diff-4c0ace26881c329c85fd42ad1a2aa6e243eb70294f2785fdbaf216f09d65668aR48

yiheng-wang-nv (approved):

Issue #395, opened 2023-04-27
Add RAM used for Bundles that use CacheDataset


yiheng-wang-nv:
Some of the bundles use CacheDataset, but do not provide any information on how many RAM are required. For a large dataset, caching all data need large RAM otherwise error will be raised. It would be hard to debug when meeting this kind of error because the error message is not informative, for example:
```
8.72s/it]WARNING:torch.distributed.elastic.multiprocessing.api:Sending process 1332 closing signal SIGTERM
ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: -9) local_rank: 0 (pid: 1331) of binary: /usr/bin/python
```

Therefore, I think we should add related information in README, and provide some suggestions for users that do not have enough RAM.

yiheng-wang-nv:
Related bundles include:

1. pathology_nuclei_classification @KumoLiu 
1. pathology_nuclei_segmentation_classification @KumoLiu 
1. endoscopic_tool_segmentation @binliunls 
1. spleen_ct_segmentation @heyufan1995 
1. spleen_deepedit_annotation @tangy5 
1. swin_unetr_btcv_segmentation @tangy5 
1. pathology_nuclick_annotation @tangy5 
1. wholeBody_ct_segmentation @tangy5 
1. pancreas_ct_dints_segmentation @dongyang0122 

May need your help to supplement the information of how many RAM is required to run the bundle training with default settings (cache rate). I personally think that it is helpful especially for bundles that train on large dataset.

For bundles that use different cache rate in `multi_gpu_train.json` and `train.json`, I think we should unify the settings to avoid complex explanations.

yiheng-wang-nv:
To sum up according to discussions as well as @zephyrie 's comments in PR: #406 #399 , I think we need to add two sections into README:

**1. (optional)  The template is like:**

### Memory Consumption

- Dataset Manager: CacheDataset
- Data Size:
- Cache Rate:
- Single GPU - System RAM Usage:
- Multi GPU (8 GPUs) - System RAM Usage:

**2. (necessary on bundles use CacheDataset) The template is like:**

### Memory Consumption Warning

If you face memory issues with CacheDataset, you can either switch to a regular Dataset class or lower the caching rate cache_rate in the configurations within range $(0, 1)$ to minimize the System RAM requirements.

Hi @zephyrie @Nic-Ma @daguangxu @dongyang0122 @tangy5 what do you think?

zephyrie:
[...]

PR #399, opened 2023-04-28
update bundle docs | four bundles

tangy5:
Related to #395 

This PR addresses the following 4 bundles:

1. spleen_deepedit_annotation 
1. swin_unetr_btcv_segmentation
1. pathology_nuclick_annotation
1. wholeBody_ct_segmentation

### DIFF
--- a/models/pathology_nuclick_annotation/docs/README.md
+++ b/models/pathology_nuclick_annotation/docs/README.md
@@ -82,9 +82,16 @@ The training was performed with the following:
 - Learning Rate: 1e-4
 - Loss: DiceLoss
 
-This bundle uses CacheDataset by default, the dataset used for training is the ConSep dataset. The dataset contains 13,136 PNG images after preprocessing
-- Single GPU: 4.7G
+### Memory Consumption
 
+- Dataset Manager: CacheDataset
+- Data Size: 13,136 PNG images
+- Cache Rate: 1.0
+- Single GPU - System RAM Usage: 4.7G
+
+### Memory Consumption Warning
+
+If you face memory issues with CacheDataset, you can either switch to a regular Dataset class or lower the caching rate `cache_rate` in the configurations within range $(0, 1)$ to minimize the System RAM requirements.
 
 ## Input
 5 channels
--- a/models/pathology_nuclick_annotation/docs/README.md
+++ b/models/pathology_nuclick_annotation/docs/README.md
@@ -82,9 +82,8 @@ The training was performed with the following:
 - Learning Rate: 1e-4
 - Loss: DiceLoss
 
-This bundle uses CacheDataset by default, the dataset used for training is the ConSep dataset.
+This bundle uses CacheDataset by default, the dataset used for training is the ConSep dataset. The dataset contains 13,136 PNG images after preprocessing
 - Single GPU: 4.7G
-- Multi GPU: 37G (8 GPUs)
 
 
 ## Input
--- a/models/pathology_nuclick_annotation/docs/README.md
+++ b/models/pathology_nuclick_annotation/docs/README.md
@@ -72,7 +72,7 @@ Example dataset.json
 }
 ```
 
-## Training configuration
+## Training Configuration
 The training was performed with the following:
 
 - GPU: at least 12GB of GPU memory
@@ -81,6 +81,10 @@ The training was performed with the following:
 - Optimizer: Adam
 - Learning Rate: 1e-4
 - Loss: DiceLoss
+
+This bundle uses CacheDataset by default, the dataset used for training is the ConSep dataset.
+- Single GPU: 4.7G
+- Multi GPU: 37G (8 GPUs)
 
 
 ## Input
--- a/models/spleen_deepedit_annotation/docs/README.md
+++ b/models/spleen_deepedit_annotation/docs/README.md
@@ -74,11 +74,16 @@ This result is benchmarked under:
  - CUDA version: 12.0
  - GPU models and configuration: A100 80G
 
-RAM consumption:
+### Memory Consumption
 
-This bundle uses CacheDataset by default, the given dataset is the MSD Task09 Spleen.
-- Single GPU: 8.2G
-- Multi GPU: 65G (8 GPUs)
+- Dataset Manager: CacheDataset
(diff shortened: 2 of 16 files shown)


models/spleen_deepedit_annotation/docs/README.md
```diff
@@ -74,6 +74,12 @@ This result is benchmarked under:
  - CUDA version: 12.0
  - GPU models and configuration: A100 80G
 
+RAM consumption:
+
+This bundle uses CacheDataset by default, the given dataset is the MSD Task09 Spleen.
```
yiheng-wang-nv: Thanks for the updates @tangy5 . Here I think we may not need to mention the dataset as it is described in `## Data`. What do you think?

tangy5: I feel it's better to add it here since the RAM consumption is closely associated with the dataset, and mentioning how many images are cached. Though we have a separate section for describing the dataset. Thanks.

Nic-Ma: Hi @zephyrie ,

Could you please help review this README doc update?

Thanks in advance.

models/pathology_nuclick_annotation/docs/README.md
```diff
@@ -82,6 +82,9 @@ The training was performed with the following:
 - Learning Rate: 1e-4
 - Loss: DiceLoss
 
+This bundle uses CacheDataset by default, the dataset used for training is the ConSep dataset. The dataset contains 13,136 PNG images after preprocessing
```
zephyrie: ### Memory Consumption
- Dataset Manager: CacheDataset
- Data: 13,136 PNG
- Cache Rate: 1.0
- Single GPU VRAM: 4.7G

models/spleen_deepedit_annotation/docs/README.md
zephyrie: Here is my suggestion:

### Memory Consumption
- Dataset Manager: CacheDataset
- Data Size: 61 3D Volumes
- Cache Rate: 1.0
- Single GPU VRAM: 8.2G
- Multi GPU VRAM: 65G (8 GPUs)

models/swin_unetr_btcv_segmentation/docs/README.md
```diff
@@ -29,6 +29,10 @@ The training as performed with the following:
 - Optimizer: Adam
 - Learning Rate: 2e-4
 
+RAM consumption:
```
zephyrie: ### Memory Consumption
- Dataset Manager: CacheDataset
- Data Size: 30 3D Volumes
- Cache Rate: 1.0
- Single GPU VRAM: 5.8G

models/wholeBody_ct_segmentation/docs/README.md
```diff
@@ -32,13 +32,28 @@ The segmentation of 104 tissues is formulated as voxel-wise multi-label segmenta
...
 
-- GPU: 32 GB of GPU memory
+- GPU: 48 GB of GPU memory
 - Actual Model Input: 96 x 96 x 96
 - AMP: True
 - Optimizer: AdamW
 - Learning Rate: 1e-4
 - Loss: DiceCELoss
 
+### GPU consumption warning:
```
zephyrie: I would move this section down to "### Latencies and memory performance of using the bundle with MONAI Label:" section with this updated text.

"### GPU Consumption Warning

Please note that the model is trained to predict 104 structures using 104 classes, which can lead to high GPU consumption when running on a single instance.

For optimal performance, refer to the benchmarking results below for the inference pipeline. Keep in mind that a CT scan with 300 slices typically requires around 27GB of memory. If you're working with larger CT scans, make sure to allocate more GPU memory or consider using the CPU for inference."

models/wholeBody_ct_segmentation/docs/README.md
```diff
@@ -32,13 +32,28 @@ The segmentation of 104 tissues is formulated as voxel-wise multi-label segmenta
...
+
+The model is trained with 104 classes in single instance, for predicting 104 structures, the GPU consumption can be large.
+
+For inference pipeline, please refer to the following section for benchmarking results. Normally, a CT scans with 300 slices will take about 27G memory, if your CT is larger, please prepare larger GPU memory or use CPU for inference.
+
+
+### RAM consumption warning:
```
zephyrie: Here is my suggestion:
"### Memory Consumption
- Dataset Manager: CacheDataset
- Data Size: 1000 3D Volumes
- Cache Rate: 0.4
- Single GPU VRAM: 83G
- Multi GPU VRAM: 666G (8 GPUs)

If you face memory issues with CacheDataset, you can either switch to a regular Dataset class or lower the caching rate `cache_rate` in the configurations (e.g., [training](../configs/train.yaml) and [multi-GPU training](../configs/multi_gpu_train.yaml)) within range $(0, 1)$ to minimize the GPU RAM requirements.
"

zephyrie:
I added suggestions for each README so that they're all mostly consistent formatting and I've commented on the DiNTS bundle review to give similar feedback.  Let me know if I didn't capture anything correctly for the sections.

models/spleen_deepedit_annotation/docs/README.md
yiheng-wang-nv: might be `RAM` rather than `GPU VRAM`

yiheng-wang-nv:
HI @zephyrie , thanks for the comments. A place to be correct is that here we should use `RAM`. `CacheDataset` cash the data into CPU memory rather than GPU memory in default.
cc @tangy5

zephyrie:
@yiheng-wang-nv would something like this make more sense?
Single GPU - System RAM Usage: 
Multi GPU (GPUs) - System RAM Usage:

yiheng-wang-nv:
> @yiheng-wang-nv would something like this make more sense? Single GPU - System RAM Usage: Multi GPU (GPUs) - System RAM Usage:

thanks @zephyrie , it looks good!

yiheng-wang-nv:
Hi @tangy5 , I will push a commit into this PR directly to address Michael's comments

yiheng-wang-nv (approved):

PR #397, opened 2023-04-27
Update `wholeBody_ct_segmentation`

KumoLiu:
### Description
While testing the lazy-resampling benchmark, I find that maybe there's room for improvement in the pre-processing. 
In this PR, I mainly update several places below:

- Remove `NormalizeIntensityd` since zeros aren't the background in the CT
- Move `ScaleIntensityd` in front of the `CropForeground` and change the value to 0-1 since we can only pad zero now in the lazy mode.
- Remove one-hot in the `AsDiscreted` which may save memory
- Update the metric in the training part from Accuracy to Dice
- Using the original label in the validation due to this [issue](https://github.com/Project-MONAI/model-zoo/issues/181#issuecomment-1510626663)

Note: the config was only tested on low resolution

cc @tangy5 @wyli 

### Status
**Work in progress**

### Please ensure all the checkboxes:

- [x] Codeformat tests passed locally by running `./runtests.sh --codeformat`.
- [ ] In-line docstrings updated.
- [ ] Update `version` and `changelog` in `metadata.json` if changing an existing bundle.
- [ ] Please ensure the naming rules in config files meet our requirements (please refer to: `CONTRIBUTING.md`).
- [ ] Ensure versions of packages such as `monai`, `pytorch` and `numpy` are correct in `metadata.json`.
- [ ] Descriptions should be consistent with the content, such as `eval_metrics` of the provided weights and TorchScript modules.
- [ ] Files larger than 25MB are excluded and replaced by providing download links in `large_file.yml`.
- [ ] Avoid using path that contains personal information within config files (such as use `/home/your_name/` for `"bundle_root"`).

PR #417, opened 2023-05-11
update whole body bundle evaluation to be GPU mem frinedy

tangy5:
Hi @yiheng-wang-nv , per QA team report, since our evaluation transforms is conducted on both prediction/labels, with 105 channels, it's hard to fit for 40G A100 gpus, 80G A100 is fine. 

Can we do this change to make evaluation mem friendly, this will make evaluation consistent with the validation during training. Then it will be good with 32/40 g GPUs for evaluatio. Inference pipeline keeps the same and is good with 32/40 GPUs already. 

Thank you.

### DIFF
--- a/models/wholeBody_ct_segmentation/configs/evaluate.json
+++ b/models/wholeBody_ct_segmentation/configs/evaluate.json
@@ -2,47 +2,23 @@
     "validate#postprocessing": {
         "_target_": "Compose",
         "transforms": [
-            {
-                "_target_": "Activationsd",
-                "keys": "pred",
-                "softmax": true
-            },
-            {
-                "_target_": "Invertd",
-                "keys": [
-                    "pred",
-                    "label"
-                ],
-                "transform": "@validate#preprocessing",
-                "orig_keys": "image",
-                "meta_key_postfix": "meta_dict",
-                "nearest_interp": [
-                    true,
-                    true
-                ],
-                "to_tensor": true
-            },
-            {
-                "_target_": "AsDiscreted",
-                "keys": [
-                    "pred",
-                    "label"
-                ],
-                "argmax": [
-                    true,
-                    false
-                ],
-                "to_onehot": 105
-            },
-            {
-                "_target_": "SaveImaged",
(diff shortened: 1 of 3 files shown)


models/wholeBody_ct_segmentation/configs/evaluate.json
```diff
@@ -7,21 +7,6 @@
                 "keys": "pred",
                 "softmax": true
             },
-            {
-                "_target_": "Invertd",
```
Nic-Ma: I thought `Invertd` is necessary for evaluation to get more accurate right?

Thanks.

KumoLiu: Maybe we could just remove `to_onehot` in `AsDiscreted` to be GPU mem friendly, just like what I do in this [PR](https://github.com/Project-MONAI/model-zoo/pull/397/files#diff-13d7daf2368303783ff1fb16c647118e08e62600c7d28f0a0102f62626ea1351).

But also need to update `MeanDice` in the train.json to be:
```
"val_mean_dice": {
                "_target_": "MeanDice",
                "include_background": false,
                "output_transform": "$monai.handlers.from_engine(['pred', 'label'])",
                "num_classes": 105    # add this line
}
```

yiheng-wang-nv: Hi @tangy5 , could you please address this question. Please also ensure the changes exactly align with eval metrics (metedata.json, README) and GPU memory (README).

tangy5: @KumoLiu @yiheng-wang-nv , thanks.

After discussing with the QA team, we will not make this evaluation change. We will keep everything as original and keep evaluation pipeline consistent. Instead, we will make more detailed documentions and isntruct users how to run low-res and high-res evaluation in a friendly way, if users doesn't has a high mem GPU, they can do low-res, or they can directly run inference instead of evaluation. 

Thanks.

yiheng-wang-nv (approved):

PR #421, opened 2023-05-13
update readme for trt memory

Can-Zhao:
Fixes # .

### Description

### Status

### DIFF
--- a/models/lung_nodule_ct_detection/docs/README.md
+++ b/models/lung_nodule_ct_detection/docs/README.md
@@ -33,7 +33,7 @@ Alternatively, we provide [resampled nifti images](https://drive.google.com/driv
 ## Training configuration
 The training was performed with the following:
 
-- GPU: at least 16GB GPU memory
+- GPU: at least 16GB GPU memory, requires 24G when exporting TRT model
 - Actual Model Input: 192 x 192 x 80
 - AMP: True
 - Optimizer: Adam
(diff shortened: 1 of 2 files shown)


models/lung_nodule_ct_detection/docs/README.md
```diff
@@ -33,7 +33,7 @@ Alternatively, we provide [resampled nifti images](https://drive.google.com/driv
 ## Training configuration
 The training was performed with the following:
 
-- GPU: at least 16GB GPU memory
+- GPU: at least 16GB GPU memory, requires 24G when exporting TRT model
```
mingxin-zheng: We hit OOM with L40G 24GB right when exporting TRT, didn't we?

Can-Zhao: I changed to 32G. I thought it means it requires 24G. So it requires 32G for TRT? What is the actual requirement?

mingxin-zheng: cc @binliunls

binliunls: Hi @mingxin-zheng ,
    I ran several tests on A100/V100. The memories are different on different GPUs. The max one is a little bit greater than 24G. So I would suggest to use GPUs, whose memory is greater than (not equal to) 24G. 32G would be a nice choice.

Thanks,
Bin

mingxin-zheng (approved):

models/lung_nodule_ct_detection/docs/README.md
mingxin-zheng: Thanks!

PR #452, opened 2023-06-08
update ldm noise scheduler api for inference

Can-Zhao:
Fixes # .

### Description
update ldm noise scheduler api for inference

### Status

### DIFF
--- a/models/brats_mri_axial_slices_generative_diffusion/configs/train_diffusion.json
+++ b/models/brats_mri_axial_slices_generative_diffusion/configs/train_diffusion.json
@@ -65,15 +65,15 @@
         "beta_start": 0.0015,
         "beta_end": 0.0195
     },
-    "inferer": {
-        "_target_": "generative.inferers.LatentDiffusionInferer",
-        "scheduler": "@noise_scheduler",
-        "scale_factor": "@scale_factor"
-    },
     "loss": {
         "_target_": "torch.nn.MSELoss"
     },
     "train": {
+        "inferer": {
+            "_target_": "generative.inferers.LatentDiffusionInferer",
+            "scheduler": "@noise_scheduler",
+            "scale_factor": "@scale_factor"
+        },
         "crop_transforms": [
             {
                 "_target_": "DivisiblePadd",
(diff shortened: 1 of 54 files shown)


yiheng-wang-nv:
Thanks @Can-Zhao , I will add the unit tests on these 2 bundles within this PR to prevent similar issues.

yiheng-wang-nv:
Hi @Nic-Ma , when preparing the unit test of ldm bundles, I just found a place of ldm bundles that is not matched up with:
https://github.com/Project-MONAI/MONAI/blob/e6ec945e4b87b90835cdad29ea64b1f27b8accda/monai/bundle/properties.py#L61

Some datasets may not have "data", such as all datasets in:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/datasets.py
or customized datasets.

Is it possible to change it into `BundleProperty.REQUIRED: False` ? Or ldm bundles should extend `BundleWorkflow` directly and define a customized workflow (within the unit test) ?

Nic-Ma:
Hi @yiheng-wang-nv ,

That's a good point, I would suggest to change the required `train_dataset_data` into `BundleProperty.REQUIRED: False` in the properties.
CC @SachidanandAlle and @tangy5 .

Thanks.

yiheng-wang-nv:
> Hi @yiheng-wang-nv ,
> 
> That's a good point, I would suggest to change the required `train_dataset_data` into `BundleProperty.REQUIRED: False` in the properties. CC @SachidanandAlle and @tangy5 .
> 
> Thanks.

I see. Hi @SachidanandAlle @tangy5 , do you think we can change the "BundleProperty.REQUIRED" of `train_dataset_data`? Are there any concerns here?

models/brats_mri_axial_slices_generative_diffusion/configs/train_diffusion.json
```diff
@@ -65,15 +65,15 @@
...
-    "inferer": {
-        "_target_": "generative.inferers.LatentDiffusionInferer",
-        "scheduler": "@noise_scheduler",
-        "scale_factor": "@scale_factor"
-    },
     "loss": {
         "_target_": "torch.nn.MSELoss"
     },
     "train": {
+        "inferer": {
```
yiheng-wang-nv, line 72: FYI @Can-Zhao , I changed to define "inferer" within "train" in order to align with:
https://github.com/Project-MONAI/MONAI/blob/858b86cf3af36e17efac8661533e3aa642dff5be/monai/bundle/properties.py#L66

yiheng-wang-nv (approved):

PR #455, opened 2023-06-13
Add more unit tests

yiheng-wang-nv:
Fixes # .

### Description
This PR is used to add unit tests on the following bundles:

- [x] spleen_deepedit_annotation
- [x] pancreas_ct_dints_segmentation
- [x] swin_unetr_btcv_segmentation
- [x] pathology_tumor_detection

### Status
**Ready**

### DIFF
--- a/ci/unit_tests/test_brats_mri_generative_diffusion.py
+++ b/ci/unit_tests/test_brats_mri_generative_diffusion.py
@@ -20,6 +20,7 @@ import nibabel as nib
 import numpy as np
 from monai.bundle import ConfigWorkflow
 from parameterized import parameterized
+from utils import check_workflow
 
 TEST_CASE_1 = [
     {
@@ -112,9 +113,7 @@ class TestLdm3d(unittest.TestCase):
             meta_file=os.path.join(bundle_root, "configs/metadata.json"),
             **override,
         )
-        trainer.initialize()
-        trainer.run()
-        trainer.finalize()
+        check_workflow(trainer, check_properties=False)
 
     @parameterized.expand([TEST_CASE_3])
     def test_autoencoder_infer(self, override):
@@ -129,9 +128,7 @@ class TestLdm3d(unittest.TestCase):
             meta_file=os.path.join(bundle_root, "configs/metadata.json"),
             **override,
         )
-        inferrer.initialize()
-        inferrer.run()
-        inferrer.finalize()
+        check_workflow(inferrer, check_properties=False)
 
     @parameterized.expand([TEST_CASE_2])
     def test_diffusion_train(self, override):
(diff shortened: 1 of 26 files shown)


models/pancreas_ct_dints_segmentation/scripts/search.py
```diff
@@ -28,7 +28,7 @@
 from monai.bundle import ConfigParser
 from monai.data import ThreadDataLoader, partition_dataset
 from monai.inferers import sliding_window_inference
-from monai.metrics import compute_meandice
+from monai.metrics import compute_dice
```
yiheng-wang-nv, line 31: update the API, now `monai.metrics` does not have `compute_meandice`
cc @dongyang0122

dongyang0122: Thanks to identify the issue! And the fix would resolve the issue.

dongyang0122 (commented):
I have reviewed the dints parts of the PR. All relevant changes are fine.

yiheng-wang-nv:
added unit tests are checked in this commit:
[727da65](https://github.com/Project-MONAI/model-zoo/pull/455/commits/727da655e1780759fb457cfb8d88dd1c083305ff)

In the latest commit, all temp changes (in order to trigger the unit tests) are reverted.

ci/unit_tests/test_pathology_tumor_detection.py
```diff
@@ -0,0 +1,113 @@
...
+
+        with open(os.path.join("models/pathology_tumor_detection", "testing.csv"), "w", newline="") as save_file:
+            wr = csv.writer(save_file)
+            wr.writerow(TIFF_INFO["csv_info"][:1])
+
+    def tearDown(self):
+        shutil.rmtree(self.dataset_dir)
+
+    @parameterized.expand([TEST_CASE_1])
+    def test_train_infer_config(self, override, override_infer):
```
KumoLiu, line 74: Looks like all these `test_train_infer_config` and `test_infer_config` are the same, do we need to extract these into functions in utils so that they can be called directly when other contributors need to write their own unit tests?

yiheng-wang-nv: Hi @KumoLiu , thanks for the suggestions. The functions are similar but may have some differences like:

1. Required files may be different, thus we still need to input many args into a utility function if defined.
2. Test ways may be different. Like in this bundle, we put train and infer together to avoid duplicating download, and in other bundles they may be in separate test functions.

Therefore, I think the code to define a config workflow should be written by contributors. However, I think we can define a function to simplify the initialize, check properties, run and finalize steps like:

```
def check_workflow(workflow: ConfigWorkflow, check_properties: bool = False):
    workflow.initialize()
    if check_properties is True:
        check_result = workflow.check_properties()
        if check_result is not None and len(check_result) > 0:
            raise ValueError(f"check properties for workflow failed: {check_result}")
    workflow.run()
    workflow.finalize()
```
Then contributor can call this function via:
```
check_workflow(trainer, False)
```

What do you think?

KumoLiu: Sounds good.

yiheng-wang-nv: Thanks, I updated in the latest commit

ci/unit_tests/test_brats_mri_generative_diffusion.py
```diff
@@ -148,14 +145,7 @@ def test_diffusion_train(self, override):
             meta_file=os.path.join(bundle_root, "configs/metadata.json"),
             **override,
         )
-        trainer.initialize()
-        # TODO: uncomment the following check after we have monai > 1.2.0
```
KumoLiu, line 152: Could you please explain why to remove these parts?

KumoLiu: Sorry, just find it in the function.

KumoLiu (approved):
Thanks for the quick update, LGTM now.

PR #463, opened 2023-07-11
474 Update model hosting place

### LINKED ISSUE #474: Update model hosting place and CI pipeline
Currently, we use Github release to store merged bundles. However, it has some limitations like rate limit.
I will modify to use NGC to store bundles to avoid some limitations.

yiheng-wang-nv:
Fixes #474 

### Description
This PR is used to update the model hosting place for bundles, it also updates the corresponding CI pipeline, Blossom will be used to replace Github workflow.

### Status
**Ready**


yiheng-wang-nv:
Hi @wyli @Nic-Ma @KumoLiu , this PR has been tested with @YanxuanLiu 's Blossom pipeline.

ci/utils.py
```diff
@@ -156,15 +163,71 @@ def get_checksum(dst_path: str, hash_func):
...
-    release_tag: str = "hosting_storage_v1",
-    repo_name: str = "Project-MONAI/model-zoo",
+    bundle_name: str,
+    version: str,
+    root_path: str,
+    bundle_zip_name: str,
+    exist_flag: bool,
+    org_name: str = "nvidia/monaihosting",
 ):
-    upload_command = f"gh release upload {release_tag} {bundle_zip_file_path} -R {repo_name}"
```
wyli, line 165: do you think it's still useful as a backup option to keep these commands?

wyli (approved):
thanks, it looks good to me, only a minor question inline

.github/workflows/update-model-info.yml
```diff
@@ -1,28 +0,0 @@
-name: update-model-info
```
wyli: maybe we should keep this for some time in case there's some issue during the transition from github to ngc

yiheng-wang-nv: Hi @wyli , thanks for the suggestion. Since `update_model_info.py` is also changed, if keep this yml file, should we create a separate update file?

wyli: sounds good @yiheng-wang-nv then in the future we can easily deprecate the github host by removing the files.

ci/utils.py
yiheng-wang-nv: Hi @wyli , same as the comments I added in the other conversation. I just pushed a new commit with unchanged files (added a `_deparate` postfix in the filename).
The old workflow will still work.

yiheng-wang-nv: As for the commands themselves, I think `gh cli` has more limitations compared with `PyGithub`. For example, it cannot support add commits, and we have to use both `git` commands and `gh cli` in order to prepare a pull request. What do you think?

.github/workflows/update-model-info.yml
yiheng-wang-nv: Good idea @wyli , maybe we can also discuss about the expected changes in MONAI core (monai.bundle.scripts), let me open a new ticket.

ci/utils.py
wyli: sure, looks like by using pygithub, our config is more readable

PR #482, opened 2023-08-05
481 rotation of output images in prostate mri anatomy

FaresAlMohamad:
### Description
481-Rotation of output images in prostate_mri_anatomy

- The transformations applied to the test data during preprocessing are not reversed during postprocessing. Thus, the resulting predictions have a different rotation compared to the original images and labels. I added an 'invertd()' function that reverses all the traceable transformations in the preprocessing.
- Running the inference on a PC without a GPU leads to an error because the handlers function 'CheckpointLoader' attempts to load the model onto the GPU. I set the map_location argument to torch.device(device) so that it automatically loads the model onto the CPU when a GPU is not available.
- Additionally, I changed some of the config directories as they were referring to the validation dataset rather than the test dataset.

### Status
Ready

### DIFF
--- a/models/prostate_mri_anatomy/configs/metadata.json
+++ b/models/prostate_mri_anatomy/configs/metadata.json
@@ -1,7 +1,8 @@
 {
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
-    "version": "0.3.2",
+    "version": "0.3.3",
     "changelog": {
+        "0.3.3": "add invertd transformation",
         "0.3.2": "add name tag",
         "0.3.1": "fix license Copyright error",
         "0.3.0": "update license files",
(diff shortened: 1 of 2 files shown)


models/prostate_mri_anatomy/configs/metadata.json
```diff
@@ -1,7 +1,8 @@
 {
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
-    "version": "0.3.2",
+    "version": "0.3.3",
     "changelog": {
+        "0.3.3": "add invertd transformation",
```
wyli, line 5: there's a problem using itk==5.2.1 https://github.com/Project-MONAI/model-zoo/pull/482/checks I think we can try to change the requirement to `"5.2"` or newer @yiheng-wang-nv

wyli (approved):

PR #491, opened 2023-08-20
Update DeepEdit infer to accept clicks from viewer

diazandr3s:
Update DeepEdit Infer config to accept clicks from the user

### DIFF
--- a/models/spleen_deepedit_annotation/configs/inference.json
+++ b/models/spleen_deepedit_annotation/configs/inference.json
@@ -86,7 +86,7 @@
                 "clip": true
             }
         ],
-        "auto_seg_transforms":[
+        "auto_seg_transforms": [
             {
                 "_target_": "Resized",
                 "keys": "image",
@@ -125,21 +125,21 @@
                 "number_intensity_ch": "@number_intensity_ch"
             }
         ],
-        "extra_transforms":[
+        "extra_transforms": [
             {
                 "_target_": "EnsureTyped",
                 "keys": "image"
             }
         ]
-        },
+    },
     "preprocessing": {
-            "_target_": "Compose",
-            "transforms": "$@infer#preprocessing_transforms + @infer#auto_seg_transforms + @infer#extra_transforms"
-        },
+        "_target_": "Compose",
+        "transforms": "$@infer#preprocessing_transforms + @infer#auto_seg_transforms + @infer#extra_transforms"
+    },
     "deepedit_preprocessing": {
-            "_target_": "Compose",
-            "transforms": "$@infer#preprocessing_transforms + @infer#deepedit_transforms + @infer#extra_transforms"
-        },
+        "_target_": "Compose",
+        "transforms": "$@infer#preprocessing_transforms + @infer#deepedit_transforms + @infer#extra_transforms"
+    },
     "dataset": {
         "_target_": "Dataset",
         "data": "$[{'image': i} for i in @datalist]",
--- a/models/spleen_deepedit_annotation/configs/inference.json
+++ b/models/spleen_deepedit_annotation/configs/inference.json
@@ -19,6 +19,7 @@
     ],
     "number_intensity_ch": 1,
     "device": "$torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')",
+    "use_click": false,
     "network_def": {
         "_target_": "DynUNet",
         "spatial_dims": 3,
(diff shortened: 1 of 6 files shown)


models/spleen_deepedit_annotation/configs/inference.json
```diff
@@ -97,13 +98,48 @@
...
             {
                 "_target_": "EnsureTyped",
                 "keys": "image"
             }
         ]
     },
+    "preprocessing": {
+        "_target_": "Compose",
+        "transforms": "$@infer#preprocessing_transforms + @infer#auto_seg_transforms + @infer#extra_transforms"
+    },
```
yiheng-wang-nv: Hi @diazandr3s , I think we can simply add one flag to control whether to use deepedit preprocessing and we may not need the "infer" section.
For example, we can define:
```
"use_deepedit": false,
```
then this section can be:
```
    "preprocessing": {
        "_target_": "Compose",
        "transforms": "$@preprocessing_transforms + (@deepedit_transforms if use_deepedit else @auto_seg_transforms) + @extra_transforms"
    },
```
and we will also not need to define `deepedit_postprocessing`

diazandr3s: Hi @yiheng-wang-nv,

Thanks for checking this out and for your feedback.
DeepEdit pre_transforms are a bit complicated. Here are the pre_transforms and the logic: https://github.com/Project-MONAI/MONAILabel/blob/main/sample-apps/radiology/lib/infers/deepedit.py#L75-L107

Also, here are the infer classes we need to define depending on the model type (segmentation/depeedit): https://github.com/Project-MONAI/MONAILabel/blob/main/sample-apps/radiology/lib/configs/deepedit.py#L143-L164

Is it possible to implement this complexity as you are suggesting with a single flag? How to implement a conditional within the inference json file.

Let me know,

yiheng-wang-nv: Hi @diazandr3s , I may make some misunderstanding here, but what I see of this bundle's changes are just sth like:
we now have "preprocessing" step which is consisted with some transforms, if we'd like to use the deepedit way, we just need to modify a part of transforms. Is it correct?

My suggestion of changes is just about how to simple this process --- using a flag to determine which set of transforms should be used, 1) current transforms 2) or replace some of them with deepedit transforms.

diazandr3s: Thanks again @yiheng-wang-nv.

I'd also like to simplify things, but it is not only defining the preprocessing. Post transforms are subject to the type of preprocessing done. Post transforms should be applied according to the applied preprocessing transforms.

In addition to this, there are some pre-transforms that are cached for faster inference: https://github.com/Project-MONAI/MONAILabel/blob/main/sample-apps/radiology/lib/infers/deepedit.py#L83C13-L83C13

We may also want to consider this for fast inference

tangy5:
HI @diazandr3s , thanks for the PR, let me also try this bundle in our MONAI Label with 3D Slicer and OHIF plugins. This bundle will be a big hit to many projects. Will comment, thank you.

models/spleen_deepedit_annotation/configs/inference.json
yiheng-wang-nv: Hi @diazandr3s , sorry I still cannot understand your point here.
In the monailabel link you posted, it also just use one flag `self.type == InferType.DEEPEDIT` to determine which set of transforms should be extended, so why can't we use the similar way in the bundle?
I do not see any blocker here, could you give an example?

Or maybe we can have a quick meeting to discuss it? cc @tangy5 @Nic-Ma

Nic-Ma:
Hi @diazandr3s ,

Please also update the README to add the expected input data format if having clicks and the CLI command to run it.

Thanks in advance.

Nic-Ma:
Hi @diazandr3s @yiheng-wang-nv ,

Thanks for the quick update.
Let's suppose to put the guidance and image together in the input dict, is it possible? Then no need to provide it in the CLI.

Thanks.

models/spleen_deepedit_annotation/docs/README.md
```diff
@@ -118,10 +118,30 @@ python -m monai.bundle run --config_file "['configs/train.json','configs/evaluat
...
+```
+python -m monai.bundle run --config_file configs/inference.json --use_click true
+```
+
+Clicks should be added to the data dictionary that is passed to the preprocessing transforms. Label names are the keys and their values are the clicks coordinates.
+
+Like the following format:
+
+```
+data_dict = {'background': [], 'spleen': [[I1, J1, K1], [I2, J2, K2]]}
```
yiheng-wang-nv: Hi @diazandr3s @Nic-Ma , for one sample, we need to prepare the data like:
```
{
    "image": "example.nii.gz",
    "background": [],
    "spleen": [[I1, J1, K1], [I2, J2, K2]],
}
```
Where the key names of "background" and "spleen" are determined by `label_name` （https://github.com/Project-MONAI/model-zoo/pull/491/files#diff-a9b28f1d091298054a8eecfef0bce115602c3259d5d2299e0a1cd34da1b5a329L11）

In order to modify the input for `dataset#data` (https://github.com/Project-MONAI/model-zoo/pull/491/files#diff-a9b28f1d091298054a8eecfef0bce115602c3259d5d2299e0a1cd34da1b5a329L109) we should design the format of the "background" and "spleen" values. The following is an example:
```
deepedit_dataset_example/
├── clicksTs
│   ├── background.npy
│   └── spleen.npy
└── imagesTs
    └── image.nii.gz
```
Then, we can modify the config file accordingly. 
I think for this specific bundle, we can use whatever format we want, and we just need to describe it clearly in the readme file (and better enhance the corresponding bundle's unit test with example data)

However, if we'd like to extend this way for more bundles, we need to carefully design a generic format. What do you think?
cc @SachidanandAlle @tangy5 @mingxin-zheng @binliunls @wyli @KumoLiu

tangy5: do we have to save the clicks to npy file? I feel this will slow down the click-based inference?

yiheng-wang-nv: If consider clicks as a kind of "labels" to be used, I think we should load them from saved files.
According to the source code of `AddGuidanceFromPointsDeepEditd ` (https://github.com/Project-MONAI/MONAI/blob/2daabf917a1e6afbc738d22db602f4859672a90d/monai/apps/deepedit/transforms.py#L680),
The input of this transform has to be a metatensor that contains keys: `image`, `background`, `spleen` and `image_meta_dict`, 

~~I'm not sure if not loading from `LoadImaged`, in other parts (of the pipeline), where can we "add" these information for each sample. What do you think?~~

Update:

I discussed with @Nic-Ma offline, instead to load from `LoadImaged`, we can input the data directly via override the `dataset#data` (https://github.com/Project-MONAI/model-zoo/pull/491/files#diff-a9b28f1d091298054a8eecfef0bce115602c3259d5d2299e0a1cd34da1b5a329L109)

for example, assume there is only one image:
```
    "dataset": {
        "_target_": "Dataset",
        "data": "[{'image': 'example1.nii.gz', 'background': [], 'spleen': [[I1, J1, K1], [I2, J2, K2]]}]",
        "transform": "@preprocessing"
    },
```

yiheng-wang-nv:
Hi @diazandr3s @tangy5 , I pushed a new commit with unit test changes and readme updates, could you please help to double confirm it? Thanks!

diazandr3s:
> Hi @diazandr3s @tangy5 , I pushed a new commit with unit test changes and readme updates, could you please help to double confirm it? Thanks!

Hi @yiheng-wang-nv,

Thanks for the update! It is working on my end.
I know MONAI Label is not a priority for this, but it is also working there. here is the PR: https://github.com/Project-MONAI/MONAILabel/pull/1527

Thanks again!

models/spleen_deepedit_annotation/docs/README.md
```diff
@@ -118,10 +118,28 @@ python -m monai.bundle run --config_file "['configs/train.json','configs/evaluat
...
+
+
+```
+python -m monai.bundle run --config_file configs/inference.json --use_click true
+```
+
+Clicks should be added to the data dictionary that is passed to the preprocessing transforms. The add keys are defined in `label_names` in `configs/inference.json`, and the corresponding values are the point coordinates. The following is an example of a data dictionary:
+
+```
+{"image": "example.nii.gz", "background": [], "spleen": [[I1, J1, K1], [I2, J2, K2]]}
```
Nic-Ma, line 139: Hi @yiheng-wang-nv ,

I feel this example may be wrong? Where is the "guidance" key?
I think it should be:
`{"image": "example.nii.gz", "guidance": {"background": [], "spleen": [[I1, J1, K1], [I2, J2, K2]]}}`
Is it right? CC @diazandr3s .

Thanks.

diazandr3s: Hi @Nic-Ma,

Sorry for the confusion. 
The **guidance** key is actually created by the second DeepEdit transform. The first DeepEdit transform searches for the label name key and its values. So it is correct as @yiheng-wang-nv has suggested:

> {"image": "example.nii.gz", "background": [], "spleen": [[I1, J1, K1], [I2, J2, K2]]}

Thanks!

tangy5:
Hi All, one more thing to think about here, if a foreground click is not associate with any organ, how can we handle the click. We will need a logic to define coordinate alone. 

how about:
{"image": "example.nii.gz", "foreground": [],  "background": [], "spleen": [[I1, J1, K1], [I2, J2, K2]]}

diazandr3s:
> Hi All, one more thing to think about here, if a foreground click is not associate with any organ, how can we handle the click. We will need a logic to define coordinate alone.
> 
> how about: {"image": "example.nii.gz", "foreground": [], "background": [], "spleen": [[I1, J1, K1], [I2, J2, K2]]}

Good point! 
However, that shouldn't be an issue in DeepEdit. This transform only reads the points for the labels defined in the configs file: https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/deepedit/transforms.py#L679

yiheng-wang-nv:
> Hi All, one more thing to think about here, if a foreground click is not associate with any organ, how can we handle the click. We will need a logic to define coordinate alone.
> 
> how about: {"image": "example.nii.gz", "foreground": [], "background": [], "spleen": [[I1, J1, K1], [I2, J2, K2]]}

Hi @tangy5 , I think for this bundle, as @diazandr3s said, since `label_name` defines it only has `background` and `spleen`, we may not need `foreground`.

models/spleen_deepedit_annotation/docs/README.md
Nic-Ma: I see, thanks!

models/spleen_deepedit_annotation/configs/inference.json
```diff
@@ -60,49 +61,79 @@
         "res_block": true
     },
     "network": "$@network_def.to(@device)",
+    "preprocessing_transforms": [
+        {
+            "_target_": "LoadImaged",
+            "keys": "image",
+            "reader": "ITKReader"
```
Nic-Ma, line 68: Hi @diazandr3s @tangy5 @yiheng-wang-nv ,

Just curious, may I know why we hard-code to use "ITKReader" here? Default to `None` doesn't work?

Thanks.

models/spleen_deepedit_annotation/configs/inference.json
```diff
@@ -60,49 +61,79 @@
...
-                "keys": "image",
-                "label_names": "@label_names",
-                "number_intensity_ch": "@number_intensity_ch"
-            },
-            {
-                "_target_": "EnsureTyped",
-                "keys": "image"
-            }
-        ]
+        "transforms": "$@preprocessing_transforms + (@deepedit_transforms if @use_click else @auto_seg_transforms) + @extra_transforms"
```
Nic-Ma, line 136: I remember @diazandr3s said we also need to add flag for the postprocessing logic, but don't see any change for it.
Is it correct?

Thanks.

models/spleen_deepedit_annotation/configs/inference.json
yiheng-wang-nv: Hi @Nic-Ma , not sure on my side, I'm also curious about why we need itkreader.

models/spleen_deepedit_annotation/configs/inference.json
diazandr3s: Hi @Nic-Ma,

Postprocessing depends on what we define as preprocessing. @yiheng-wang-nv proposed a simpler solution to this which works as expected.

Thanks!

models/spleen_deepedit_annotation/configs/inference.json
diazandr3s: This DeepEdit version works specifically on 3D images (CT and MR images). I've found a significant amount of NIfTI images with corrupted headers - they don't work with other readers. 
We can change the reader, this is just a heads-up if we change the reader type.

Nic-Ma: OK, I see, let's keep it for now. thanks.

Nic-Ma (approved):
Looks good to me now.
@yiheng-wang-nv Do you have any other comments?
If not, let's merge it.

Thanks.

PR #504, opened 2023-09-14
Enhance Deepedit bundle to enable finetune and early stopping

yiheng-wang-nv:
### Description
This PR is used to enable early stopping and finetune for the spleen deepedit annotation bundle.

### Status
**Ready**

### DIFF
--- a/models/spleen_deepedit_annotation/configs/multi_gpu_train.json
+++ b/models/spleen_deepedit_annotation/configs/multi_gpu_train.json
@@ -31,6 +31,7 @@
         "$monai.utils.set_determinism(seed=123)"
     ],
     "run": [
+        "$@validate#handlers#0.set_trainer(trainer=@train#trainer) if @early_stop else None",
         "$@train#trainer.run()"
     ],
     "finalize": [
--- a/models/spleen_deepedit_annotation/configs/multi_gpu_train.json
+++ b/models/spleen_deepedit_annotation/configs/multi_gpu_train.json
@@ -23,7 +23,37 @@
         "shuffle": false
     },
     "validate#dataloader#sampler": "@validate#sampler",
-    "validate#evaluator#val_handlers": "$None if dist.get_rank() > 0 else @validate#handlers",
+    "validate#handlers": [
+        {
+            "_target_": "EarlyStopHandler",
+            "_disabled_": "$not @early_stop",
+            "trainer": null,
+            "patience": 1,
+            "score_function": "$scripts.score_function",
+            "min_delta": 0.01
+        },
+        {
+            "_target_": "StatsHandler",
+            "_disabled_": "$dist.get_rank() > 0",
+            "iteration_log": false
+        },
+        {
+            "_target_": "TensorBoardStatsHandler",
+            "_disabled_": "$dist.get_rank() > 0",
+            "log_dir": "@output_dir",
+            "iteration_log": false
+        },
+        {
+            "_target_": "CheckpointSaver",
+            "_disabled_": "$dist.get_rank() > 0",
+            "save_dir": "@ckpt_dir",
+            "save_dict": {
+                "model": "@network"
+            },
+            "save_key_metric": true,
+            "key_metric_filename": "model.pt"
+        }
+    ],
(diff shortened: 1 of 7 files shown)


models/spleen_deepedit_annotation/configs/multi_gpu_train.json
```diff
@@ -31,6 +31,7 @@
         "$monai.utils.set_determinism(seed=123)"
     ],
     "run": [
+        "$@validate#handlers#0.set_trainer(trainer=@train#trainer) if @early_stop else None",
```
yiheng-wang-nv, line 34: Hi @wyli @Nic-Ma , do you know if we can use early stopping with distributed training？
I tried with this config but the program hangs in `finalize` (when execute `dist.destroy_process_group()`).

wyli: fyi https://github.com/pytorch/ignite/pull/666

models/spleen_deepedit_annotation/configs/multi_gpu_train.json
```diff
@@ -23,14 +23,45 @@
         "shuffle": false
     },
     "validate#dataloader#sampler": "@validate#sampler",
-    "validate#evaluator#val_handlers": "$None if dist.get_rank() > 0 else @validate#handlers",
+    "validate#handlers": [
+        {
+            "_target_": "EarlyStopHandler",
```
yiheng-wang-nv: this handler should run on all ranks.

models/spleen_deepedit_annotation/configs/multi_gpu_train.json
wyli: perhaps we can enhance the docstring here for future references https://github.com/Project-MONAI/MONAI/blob/6dfe17a7bfda9b4ddfa19b60886b7fa14e816afd/monai/handlers/earlystop_handler.py#L50-L51

models/spleen_deepedit_annotation/configs/multi_gpu_train.json
```diff
@@ -23,14 +23,45 @@
...
+            "_target_": "CheckpointSaver",
+            "_disabled_": "$dist.get_rank() > 0",
+            "save_dir": "@ckpt_dir",
+            "save_dict": {
+                "model": "@network"
+            },
+            "save_key_metric": true,
+            "key_metric_filename": "model.pt"
+        }
+    ],
```
Nic-Ma: I think you don't need to put all the handlers here, can you try:
```json
"validate#evaluator#val_handlers": "$@validate#handlers[: 1 if dist.get_rank() > 0 else None]",
```

yiheng-wang-nv: Hi @Nic-Ma , if not using early stopping, `validate#handlers[: 1]` will be `StatsHandler` thus there will have duplicate logs.
Here the disable way is different between `EarlyStopHandler ` and other handlers, thus I did not use one line code to override. What do you think?

yiheng-wang-nv: Hi @Nic-Ma , in this way when `early_stop` is False, `StatsHandler` will be used on all ranks.

yiheng-wang-nv: I added a flag to enable this way, thanks @Nic-Ma

Nic-Ma: If the early_stop is not sure, you can use [: -2] instead of [: 1] to avoid this complicated logic, refer to the train part:
https://github.com/Project-MONAI/model-zoo/blob/dev/models/spleen_deepedit_annotation/configs/multi_gpu_train.json#L18

Thanks.

yiheng-wang-nv: Thanks Nic, it's better to use negative index. I changed to `[: -3]` since we only need the first handler.

Nic-Ma (approved):

PR #514, opened 2023-10-06
Segment Template

### LINKED ISSUE #513: Segmentation Template For the Model Zoo

ericspod:
Fixes #513 .

### Description
This is the template for segmentation bundles. Users can create their own new bundles by copying and adapting this one. 

Multi-GPU support is a work in progress, it doesn't work for me but it should be fine.

### Status
**Ready**


wyli (approved):
thanks, it looks good to me

models/segmentation_template/docs/README.md
```diff
@@ -0,0 +1,75 @@
+
+# Template Segmentation Bundle
+
+This bundle is meant to be an example of segmentation in 3D which you can copy and modify to create your own bundle.
```
yiheng-wang-nv, line 4: A good template. Should we also modify the readme of the repo to mention it, thus users can see it directly?

ericspod: I've added a mention in the README, have a look but it's a section we can expand with new templates as we go.

models/segmentation_template/configs/test.yaml
```diff
@@ -0,0 +1,125 @@
...
+  key_val_metric:
+    val_mean_dice:
+      _target_: MeanDice
+      include_background: false
+      output_transform: $monai.handlers.from_engine([@pred, @label])
+  val_handlers: '@handlers'
+
+run:
+- $@evaluator.run()
+- '$print(''Per-image Dice:\n'',@evaluator.state.metric_details[''val_mean_dice''].cpu().numpy())'
```
yiheng-wang-nv, line 125: is it a typo here? I think the comma used here has invalid syntax

ericspod: I'm pretty sure this is correct, without YAML formatting and bundle sigils it's equivalent to this Python code:

```python
print('Per-image Dice:\n',evaluator.state.metric_details['val_mean_dice'].cpu().numpy())
```

yiheng-wang-nv:
We may also think about how to check if the whole .yaml files do not have typo or syntax issues. It seems we cannot use similar unit tests on this template bundle.
cc @wyli @KumoLiu @Nic-Ma

models/segmentation_template/configs/test.yaml
yiheng-wang-nv: Oh sorry, it's yaml's format. Thanks @ericspod for the explain.

ericspod:
> We may also think about how to check if the whole .yaml files do not have typo or syntax issues. It seems we cannot use similar unit tests on this template bundle.

For this we can simply run any YAML file through `pyyaml`, any JSON file through the in-built `json` library, and then further check any Python expressions by attempting to parse them with the in-built `ast` library. This would do a static check without running the code that should be good enough to catch most things.

PR #533, opened 2023-11-29
Add classification template

KumoLiu:
Part of tutorial#1456

### Description
Add a classification template

### Status
**Ready**

### DIFF
--- a/models/classification_template/configs/inference.yaml
+++ b/models/classification_template/configs/inference.yaml
@@ -1,0 +1,115 @@
+# This implements the workflow for applying the network to a directory of images and measuring network performance with metrics.
+
+imports:
+- $import os
+- $import json
+- $import torch
+- $import glob
+
+# pull out some constants from MONAI
+image: $monai.utils.CommonKeys.IMAGE
+label: $monai.utils.CommonKeys.LABEL
+pred: $monai.utils.CommonKeys.PRED
+
+# hyperparameters for you to modify on the command line
+batch_size: 1  # number of images per batch
+num_workers: 0  # number of workers to generate batches with
+num_classes: 4  # number of classes in training data which network should predict
+device: $torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')
+
+# define various paths
+bundle_root: .  # root directory of the bundle
+ckpt_path: $@bundle_root + '/models/model.pt'  # checkpoint to load before starting
+dataset_dir: $@bundle_root + '/data/test_data'  # where data is coming from
+
+# network definition, this could be parameterised by pre-defined values or on the command line
+network_def:
+  _target_: DenseNet121
+  spatial_dims: 2
+  in_channels: 1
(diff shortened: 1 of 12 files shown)


models/classification_template/configs/inference.yaml
```diff
@@ -0,0 +1,115 @@
...
+# transform to apply to data from network to be suitable for validation
+postprocessing:
+  _target_: Compose
+  transforms:
+  - _target_: Activationsd
+    keys: '@pred'
+    softmax: true
+  - _target_: AsDiscreted
+    keys: ['@pred', '@label']
+    argmax: [true, false]
```
ericspod, line 76:
```suggestion
  - _target_: AsDiscreted
    keys: '@pred'
    argmax: true
```

I don't think we want to apply anything to the label here?

models/classification_template/configs/train.yaml
```diff
@@ -0,0 +1,233 @@
...
+# transform to apply to data from network to be suitable for validation
+postprocessing:
+  _target_: Compose
+  transforms:
+  - _target_: Activationsd
+    keys: '@pred'
+    softmax: true
+  - _target_: AsDiscreted
+    keys: ['@pred', '@label']
+    argmax: [true, false]
```
ericspod, line 141: Same thing I suggested above.

ericspod:
Hi @KumoLiu looks good! I had a few points about how transforms are being applied but it works for you so it's just a best practice thing. One other suggestion is to have a `common.yaml` file with all the common definitions in it (eg. `network_def`), then include that when doing things like training. I hadn't done that for the segmentation template because I wanted to maintain compatibility with the current version of MONAILabel. Here we don't need to worry about that so we can change things now to have a common file or leave it for a future refactor.

models/classification_template/configs/inference.yaml
KumoLiu: I added this one here to one hot label that may used in `ConfusionMetrix`.

KumoLiu:
Hi @ericspod, I didn't get the point of why we need the `common.yaml`. I see you created an issue [here](https://github.com/Project-MONAI/MONAILabel/issues/1549) and also mentioned it in the dev meeting. But seems we don't need to specify the common field in multiple configs, we only need to add what we want to override in the `evaluate.yaml` or `inference.yaml` or other configs.

KumoLiu:
After reading again the configs, I think I get the points. It makes sense if we add a `common.yaml` but every time we may need to specify multiple configs due to the existence of `common.yaml`.
Perhaps we could refactor it later when we reach a consensus. What do you think?

ericspod:
I'm good to refactor later, it's something we will likely do for both template bundles. I had recommended having a common config file in [this bundle tutorial](https://github.com/Project-MONAI/tutorials/blob/main/bundle/03_mednist_classification_v2.ipynb) but it's not necessary.

ericspod (approved):
LGTM!

PR #620, opened 2024-08-16
Implementation of TRT wrapping via inference.json

borisfom:
Sample implementation of using new trt_wrap() from MONAI
Depends on https://github.com/Project-MONAI/MONAI/pull/7990


borisfom:
I observe 2.5x speedup with FP16 TRT for a small sample (28.57it/s vs 11it/s). Did not check the accuracy.

models/pathology_nuclei_classification/configs/inference_trt.json
```diff
@@ -1,12 +0,0 @@
-{
```
Nic-Ma, line 1: I would suggest to keep this `inference_trt.json` file and put all your `trt_wrap` logic here.
Then users still run the previous behavior:
`python -m monai.bundle run --config_file "['configs/inference.json', 'configs/inference_trt.json']"`
It will automatically merge these 2 JSON files:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/config_parser.py#L425
What do you think?

Thanks.

models/brats_mri_generative_diffusion/configs/inference.json
```diff
@@ -63,12 +65,20 @@
         ],
         "num_res_blocks": 2
     },
+    "trt_args": {
```
Nic-Ma: I see every bundle in this PR contain the same `trt_args`, can we set some default values for `trt_args` in the `trt_wrap` function? Then no need to write this part for every bundle.

Thanks.

models/brats_mri_generative_diffusion/configs/inference.json
```diff
@@ -2,13 +2,15 @@
...
-        "$from pathlib import Path"
+        "$from pathlib import Path",
+        "$from monai.networks import trt_wrap"
     ],
     "bundle_root": ".",
     "model_dir": "$@bundle_root + '/models'",
     "output_dir": "$@bundle_root + '/output'",
     "create_output_dir": "$Path(@output_dir).mkdir(exist_ok=True)",
     "device": "$torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')",
+    "trt": "$True",
```
Nic-Ma: If keep the `inference_trt.json`, then don't need this flag anymore.

Thanks.

models/pathology_nuclei_classification/configs/inference_trt.json
borisfom: Well, from the developer perspective, this is kind of clean unintrusive way to specify TRT support, and most of inference_trt.json are going to be identical:
... imports and trt_args here ... 
"network": "$trt_wrap(@network_def.to(@device), @checkpoint_path, @trt_args)"

From the user perspective,  especially if TRT-accelerated inference is a default, it may be simpler to have one config and --trt switch.  

I am fine with either approach.

models/brats_mri_generative_diffusion/configs/inference.json
borisfom: Yes I was thinking about this, too. Let me take another stab at this.

models/brats_mri_generative_diffusion/configs/inference.json
borisfom: Well yes I did not realize you support merging of the configs - let's keep this discussion in the thread above.

models/pathology_nuclei_classification/configs/inference_trt.json
Nic-Ma: Thanks @borisfom .
MONAI is a community OSS project, not only supporting NVIDIA GPUs, so we don't enable TensorRT by default, `inference_trt.json` can help isolate this advanced acceleration feature.
What do you think?

Thanks.

borisfom: If it's not on by default, then from user perspective separate inference_trt.json is close to --trt flag, and better in terms of source structure. Let's go with that, sure.

models/brats_mri_generative_diffusion/configs/inference.json
borisfom: This is implemented now.

models/vista3d/configs/inference_trt.json
```diff
@@ -0,0 +1,12 @@
...
+    "imports": [
+        "$import glob",
+        "$import json",
+        "$import pathlib",
+        "$import os",
+        "$import scripts",
+        "$import numpy as np",
+        "$from monai.networks import trt_wrap"
+    ],
+    "network": "$trt_wrap(@network_def.to(@device), @bundle_root + '/models/model.pt', submodule=['image_encoder.encoder', 'class_head'])"
```
yiheng-wang-nv: thanks for the changes.
~~Currently model zoo CI excludes vista3d bundle's torchscript test.
After trt_wrap merging into MONAI, I can modify: https://github.com/Project-MONAI/model-zoo/blob/dev/ci/verify_bundle.py#L339 to enable it.~~

Update: I misunderstood here. The place to change is the `verify_tensorrt.py` file

models/brats_mri_generative_diffusion/configs/metadata.json
```diff
@@ -13,7 +14,7 @@
         "1.0.1": "update dependency, update trained model weights",
         "1.0.0": "Initial release"
     },
-    "monai_version": "1.3.1",
+    "monai_version": "1.3.3rc1",
```
Nic-Ma: @KumoLiu I think we should use `1.4rc` here?

Thanks.

Nic-Ma (approved):
Hi @borisfom ,

Thanks for the update, it overall looks good to me.
I think you need to update the MONAI version in `metadata.json` of every bundle? CC @yiheng-wang-nv .

Thanks.

models/brats_mri_generative_diffusion/configs/metadata.json
KumoLiu: should use `1.4` here.

KumoLiu:
Hi @borisfom, Did you verify if the performance is comparable to the previous version in these models and if the inference time has been reduced as expected? I recall that we have a place to document these numbers, though I can't remember exactly where—perhaps in the README or somewhere else. @binliunls, could you please help with this verification? Thanks.

yiheng-wang-nv:
> Hi @borisfom, Did you verify if the performance is comparable to the previous version in these models and if the inference time has been reduced as expected? I recall that we have a place to document these numbers, though I can't remember exactly where—perhaps in the README or somewhere else. @binliunls, could you please help with this verification? Thanks.

Hi @KumoLiu @borisfom , in README, like:
https://github.com/Project-MONAI/model-zoo/tree/dev/models/spleen_ct_segmentation#tensorrt-speedup

yiheng-wang-nv:
Hi @borisfom , your vista2d, vista3d PR is merged, could you resolve the conflicts and then we can push this PR forward, thanks!

Nic-Ma:
Hi @binliunls ,

Could you please help verify this PR locally?
If no issues and TensorRT does improve the inference speed, let's merge it ASAP and include in the release.
CC @KumoLiu , @yiheng-wang-nv .

Thanks in advance.

binliunls:
> Hi @binliunls ,
> 
> Could you please help verify this PR locally? If no issues and TensorRT does improve the inference speed, let's merge it ASAP and include in the release. CC @KumoLiu , @yiheng-wang-nv .
> 
> Thanks in advance.

I am working on it. All the benchmark results will go to [this PR](https://github.com/Project-MONAI/model-zoo/pull/640). However for the diffusion models like `brats_mri_generative_diffusion` and `brats_mri_axial_slices_generative_diffusion`, I don't think the update in this PR is completed. @borisfom  please complete them or add them in another PR if these bundles are not supported for now.

Thanks,  
Bin

borisfom:
@binliunls : I have merged origin/dev and fixed brats_mri_generative_diffusion and brats_mri_axial_slices_generative_diffusion. Note: for brats_mri_axial_slices_generative_diffusion, default 
python -m monai.bundle run --config_file "['configs/inference.json', 'configs/inference_trt.json']" does not get to the diffusion forward, so no TRT is generated, and there may be issues with it. Let me know how to run it so that the diffusion actually gets executed.

binliunls:
> @binliunls : I have merged origin/dev and fixed brats_mri_generative_diffusion and brats_mri_axial_slices_generative_diffusion. Note: for brats_mri_axial_slices_generative_diffusion, default python -m monai.bundle run --config_file "['configs/inference.json', 'configs/inference_trt.json']" does not get to the diffusion forward, so no TRT is generated, and there may be issues with it. Let me know how to run it so that the diffusion actually gets executed.

Hi @borisfom , 
    Thank you for your update. As shown in the inference config file, you should compile the [diffusion](https://github.com/Project-MONAI/model-zoo/blob/ae0beb1049fffd790ecd537dee4af2db7dfe89fb/models/brats_mri_generative_diffusion/configs/inference.json#L74) and [autoencoder](https://github.com/Project-MONAI/model-zoo/blob/ae0beb1049fffd790ecd537dee4af2db7dfe89fb/models/brats_mri_generative_diffusion/configs/inference.json#L71) with TRT instead of network, since the `network` parameter is not defined in either bundle. Please check if you can compile both `diffusion` and `autoencoder`. 

Thanks,  
Bin

borisfom:
@binliunls : Thanks, you are correct! Works as expected now.

yiheng-wang-nv (approved):

PR #625, opened 2024-08-21
Update brats_mri_generative_diffusion to monai1.4

Can-Zhao:
Fixes # .

### Description

### Status


models/brats_mri_generative_diffusion/configs/metadata.json
```diff
@@ -1,7 +1,8 @@
 {
-    "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_generator_ldm_20230507.json",
-    "version": "1.0.9",
+    "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20240725.json",
+    "version": "1.0.10",
```
KumoLiu: I guess here should be "1.1.0"?

Can-Zhao: I was thinking we would change it to 1.1.0 when we switch to new model weights in the future? I'm fine with either way

yiheng-wang-nv (approved):

yiheng-wang-nv:
Hi @Can-Zhao , the changes look good to me. Have you verified the performance after loading old weights?

Can-Zhao:
> Hi @Can-Zhao , the changes look good to me. Have you verified the performance after loading old weights?

I checked the outputs of inference.json

yiheng-wang-nv:
Hi @Can-Zhao , could you please update the autoencoder's torchscript model? CI shows it's weight keys mismatch: https://github.com/Project-MONAI/model-zoo/actions/runs/10556654519/job/29242633314

models/brats_mri_generative_diffusion/configs/metadata.json
KumoLiu: After discussed offline with @Nic-Ma, let's use `1.1.0` here since we removed generative dependency in this version.

Nic-Ma: Yes, it should be a major update, thanks.

Can-Zhao: Sure, will do. Thanks!

Can-Zhao: Done, thanks

yiheng-wang-nv:
```
torch.jit.frontend.NotSupportedError: Compiled functions can't take variable number of arguments or use keyword-only arguments with defaults:
  File "/home/venn/Desktop/monai-code/monai_service_env/lib/python3.10/site-packages/torch/_compile.py", line 343
    determinism_check: str = _DEFAULT_DETERMINISM_MODE,
    debug: bool = False,
    **kwargs
     ~~~~~~~ <--- HERE
):
    r"""Checkpoint a model or part of the model.
'AutoencoderKL.encode' is being compiled since it was called from 'AutoencoderKL.forward'
  File "/home/venn/Desktop/monai-code/MONAI/monai/networks/nets/autoencoderkl.py", line 664
    def forward(self, x: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
        z_mu, z_sigma = self.encode(x)
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
        z = self.sampling(z_mu, z_sigma)
        reconstruction = self.decode(z)

```

Hi @Can-Zhao @KumoLiu @ericspod , I pushed a new commit to remove torchscript check for auto encoderkl, since it has errors shown above. Is it expected?

KumoLiu:
Hi @yiheng-wang-nv @ericspod and @Can-Zhao, I've checked and found that the last version where AutoEncoderKL could successfully be converted to TorchScript was with monai-generative 0.2.2, which didn't include the checkpoint option. It seems that due to the changes made after this version, AutoEncoderKL is no longer compatible with TorchScript conversion.

I'm fine with moving this check if @ericspod doesn't have any other concerns. Thanks.

ericspod:
> Hi @yiheng-wang-nv @ericspod and @Can-Zhao, I've checked and found that the last version where AutoEncoderKL could successfully be converted to TorchScript was with monai-generative 0.2.2, which didn't include the checkpoint option. It seems that due to the changes made after this version, AutoEncoderKL is no longer compatible with TorchScript conversion.
> 
> I'm fine with moving this check if @ericspod doesn't have any other concerns. Thanks.

It may be a versioning issue with the combination of Python and Pytorch that you or @Can-Zhao is using. With the current version of GenerativeModels, Python 3.12, Pytorch 2.4, MONAI 1.3.3rc1, I can add the following to the end of `test_autoencoderkl.py` (in `GenerativeModels` not MONAI) and it passes:

```python
    @parameterized.expand(list(product([True, False], repeat=2)))
    def test_torchscript(self, use_checkpointing, use_convtranspose):
        from tests.utils import test_script_save
        input_param = {
            "spatial_dims": 2,
            "in_channels": 1,
            "out_channels": 1,
            "num_channels": (4, 4, 4),
            "latent_channels": 4,
            "attention_levels": (False, False, False),
            "num_res_blocks": 1,
            "norm_num_groups": 4,
            "use_checkpointing": use_checkpointing,
            "use_convtranspose": use_convtranspose,
        }
        net = AutoencoderKL(**input_param).to(device)
        test_script_save(net, torch.rand(1, 1, 16, 16))
```

This checks every combination of checkpointing and transpose being on or off and I get no errors. Does this test miss something you're seeing?

KumoLiu:
> It may be a versioning issue with the combination of Python and Pytorch that you or @Can-Zhao is using. With the current version of GenerativeModels, Python 3.12, Pytorch 2.4, MONAI 1.3.3rc1, I can add the following to the end of `test_autoencoderkl.py` (in `GenerativeModels` not MONAI) and it passes:

I tried with the version in the monai, it can't convert correctly. Do you think it's expected? Seems Rearrange can not be compatible with the torchscript convert.

ericspod:
> I tried with the version in the monai, it can't convert correctly. Do you think it's expected? Seems Rearrange can not be compatible with the torchscript convert.

I made up a simple test with `Rearrange` that worked so this doesn't seem to be the issue, or at least not always. I tried to save an instance of `monai.networks.nets.autoencoderkl.Encoder` and this triggered a complaint for `Rearrange`, saying it didn't have a `__init__`. 

```python
e=Encoder(2,1,[8,16],1,[2,2],1,1e5,[True,True])
torch.jit.script(e).save("/tmp/tmp.pt")
```
From this I get:
```
RuntimeError: 
Class Rearrange does not have an __init__ function defined:
  File "/home/localek10/workspace/monai/MONAI_mine/monai/networks/blocks/spatialattention.py", line 80
        if self.spatial_dims == 1:
            h = x.shape[2]
            rearrange_input = Rearrange("b c h -> b h c")
                              ~~~~~~~~~ <--- HERE
            rearrange_output = Rearrange("b h c -> b c h", h=h)
        if self.spatial_dims == 2:
```
I think this is something that's expected currently as we haven't narrowed down what the issue is with out version. Testing Torchscript failures has been difficult and muddled by the caching that happens in the background. I've had some failed tests complain about trying to redefine methods as a result, I'm working on how to get around that. Can we just work around this for this PR?

KumoLiu:
> Can we just work around this for this PR?

Sure, let's move torchscript convert to this version and revisit it later when we have a solution or identify the underlying cause. @yiheng-wang-nv

PR #628, opened 2024-08-23
Updated vista2d to use monai components

myron:
Update to vista2d to use net from monai.


KumoLiu (commented):
Hi @myron, thanks for the PR. Looks good to me.
Please also update the metadata.json. Thanks!

yiheng-wang-nv (approved):

PR #650, opened 2024-09-11
642 (part 2) update large files for a part of bundles

yiheng-wang-nv:
Fixes # .

### Description

### Status


models/mednist_reg/configs/metadata.json
```diff
@@ -1,7 +1,8 @@
 {
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_20220324.json",
-    "version": "0.0.4",
+    "version": "0.0.5",
     "changelog": {
```
KumoLiu, line 4: Do we need update the monai version for this bundle?

KumoLiu (approved):
Thanks for the update, overall looks good to me.

models/mednist_reg/configs/metadata.json
yiheng-wang-nv: I think we can keep the old version. The reason why I update for some other bundles is that our CI uses python 3.10 and cannot support old torch versions.

KumoLiu: Ok, but I guess the ci have already test with the latest monai version, so the bundle can work with the new version, right?

yiheng-wang-nv: for each bundle, CI uses it's corresponding monai/torch/... versions.

PR #654, opened 2024-09-16
Fix the bug that MAISI ckpt cannot be loaded after finetune. 

guopengf:
Fixes # .

### Description

MAISI output checkpoint after finetuning cannot be used as ``trained_controlnet_path``.
This problem came from the `CheckpointSaver`. When a single key is provided for `save_dict` such as:
"save_dict": {
                    "controlnet_state_dict": "@controlnet"
 },
The saved dict does contain the key " "controlnet_state_dict". However, it directly saves the state_dict of controlnet as the checkpoint. 

The workaround is that we also save the optimizer state. For example,
"save_dict": {
                    "controlnet_state_dict": "@controlnet",
                    "optimizer": "@optimizer"
 }. Then, the MAISI output checkpoint after fine-tuning can be properly loaded.

### Status


KumoLiu (approved):
Thanks for the fix, looks good to me.

PR #665, opened 2024-09-20
Remove download large file and update readme

KumoLiu:
### Description
- Remove download large files to avoid confusion as the large files will be directly downloaded when download the bundle.
- Set weights_only arg during torch.load to avoid bunch of warning.
- Modify error message to make it clearer.

### Status
**Ready**


models/maisi_ct_generative/configs/metadata.json
```diff
@@ -1,7 +1,8 @@
 {
     "schema": "https://github.com/Project-MONAI/MONAI-extra-test-data/releases/download/0.8.1/meta_schema_generator_ldm_20240318.json",
-    "version": "0.4.2",
+    "version": "0.4.3",
     "changelog": {
+        "0.4.3": "remove download large files",
```
dongyang0122: Please also add texts like "adding weights_only=False in the model weight loading" in this line. Thanks.

models/maisi_ct_generative/scripts/sample.py
```diff
@@ -923,7 +923,7 @@ def ensure_output_size_and_spacing(self, labels, check_contains_target_labels=Tr
                         raise ValueError(
                             (
                                 f"Resampled mask does not contain required class labels {anatomy_label}. "
-                                "Please tune spacing and output size."
+                                "Please consider increasing the spacing or specifying a larger output size."
```
dongyang0122: "increase the output spacing"

models/maisi_ct_generative/docs/README.md
```diff
@@ -23,10 +23,6 @@ The inference requires:
 - Disk Memory: at least 21GB disk memory
 
 ### Execute inference
-The model weights can be downloaded with
-```
-python -m scripts.download_files
-```
```
dongyang0122, line 29: I am not sure how users can download the model weights if we remove this part.

models/maisi_ct_generative/configs/metadata.json
KumoLiu: Hi @dongyang0122, I update this part to "add weights_only when loading weights" since I didn't set all of them to False, only two of them include other parameter info so I set to False.

models/maisi_ct_generative/scripts/sample.py
KumoLiu: Thanks, update this part in the latest commit.

models/maisi_ct_generative/docs/README.md
KumoLiu: Hi @dongyang0122, users don't need download large files by themselves since when they downloading the bundle, all these large files have already been downloaded and inlcuded in the bundle.

dongyang0122 (approved):

PR #674, opened 2024-09-26
Fix vista3d bundle evaluation bug

heyufan1995:
Fixes # .

### Description
Fix the evaluation bug.


KumoLiu (approved):
Should also update metadata.json
Overall looks good to me, thanks.

yiheng-wang-nv:
I pushed minor changes on readme, and confirmed that the eval config gets metric ~0.96

yiheng-wang-nv (approved):

KumoLiu:
https://github.com/Project-MONAI/model-zoo/blob/dev/models/vista3d/scripts/evaluator.py#L18
https://github.com/Project-MONAI/model-zoo/blob/dev/models/vista3d/scripts/trainer.py#L19
Here need to be updated to `from monai.utils import IgniteInfo`

PR #692, opened 2024-10-08
Update readme in vista3d

KumoLiu:
Update readme in vista3d regarding if input data missing header information 

### Status

### DIFF
--- a/models/vista3d/docs/README.md
+++ b/models/vista3d/docs/README.md
@@ -116,7 +116,8 @@ The default configs for both variables are derived from the `label_mappings` con
 "val_label_set": "$[0] + list(x[0] for x in @label_mappings#default)"
 ```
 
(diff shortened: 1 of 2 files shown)


models/vista3d/docs/README.md
```diff
@@ -116,6 +116,8 @@ The default configs for both variables are derived from the `label_mappings` con
 "val_label_set": "$[0] + list(x[0] for x in @label_mappings#default)"
 ```
 
+If the user's data lacks header information, MONAI will provide some default header information. After inference, the user may need to apply some transforms to the data to match the original image, or they can add the correct header information to the input data themselves.
```
yiheng-wang-nv: what does the "data" here mean? What kind of "header"?
I think the sentences here are not clear, and suggest to give some code example

yiheng-wang-nv (approved):

PR #743, opened 2025-04-08
[part 2] Update all bundles to Huggingface and fix missing dependencies

### LINKED ISSUE #739: Update all bundles to Huggingface
Based on the host change in: https://github.com/Project-MONAI/model-zoo/pull/737
We can update a new version on all bundles to make sure that they are in Huggingface. It's helpful to redesign the  download API.

yiheng-wang-nv:
Fixes #739 .

### Description

### Status
**Ready**

### DIFF
--- a/ci/get_bundle_requirements.py
+++ b/ci/get_bundle_requirements.py
@@ -19,8 +19,6 @@ from utils import get_json_dict
 
 ALLOW_MONAI_RC = os.environ.get("ALLOW_MONAI_RC", "false").lower() in ("true", "1", "t", "y", "yes")
 
-SPECIAL_LIB_LIST = ["xformers"]
-
 
 def increment_version(version):
     """
--- a/ci/get_bundle_requirements.py
+++ b/ci/get_bundle_requirements.py
@@ -81,12 +81,7 @@ def get_requirements(bundle, models_path, requirements_file):
             if package_key in metadata.keys():
                 optional_dict = metadata[package_key]
                 for name, version in optional_dict.items():
-                    if name not in SPECIAL_LIB_LIST:
-                        libs.append(f"{name}=={version}")
-                    else:
-                        if "pytorch_version" in metadata.keys():
-                            # remove torch from libs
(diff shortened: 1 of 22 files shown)


yiheng-wang-nv:
> Copilot reviewed 3 out of 22 changed files in this pull request and generated no comments.
> 
> Files not reviewed (19)
> Comments suppressed due to low confidence (2)

`xformers` is no need to be installed due to maisi related changes in MONAI, thus it's removed.

ci/get_bundle_requirements.py
```diff
@@ -81,12 +79,7 @@ def get_requirements(bundle, models_path, requirements_file):
             if package_key in metadata.keys():
                 optional_dict = metadata[package_key]
                 for name, version in optional_dict.items():
-                    if name not in SPECIAL_LIB_LIST:
-                        libs.append(f"{name}=={version}")
-                    else:
-                        if "pytorch_version" in metadata.keys():
```
KumoLiu, line 87: What's the reason for this change?

models/pancreas_ct_dints_segmentation/configs/metadata.json
```diff
@@ -31,12 +32,15 @@
     "monai_version": "1.4.0",
```
KumoLiu, line 32: Do we need update the monai version when monai tagged 1.5?

ci/get_bundle_requirements.py
yiheng-wang-nv: when we created the initial maisi bundle, it requires xformers which is not included in MONAI, thus we need to install it when running CI, and xformers cannot be simply installed via pip command, thus we add some other code to handle it

models/pancreas_ct_dints_segmentation/configs/metadata.json
yiheng-wang-nv: HI @KumoLiu , it depends on the users, there is no hard requirements. However, if a bundle is too old (for example, uses a torch version that is not supported by our CI's python version), we may update it.

KumoLiu (approved):

Issue #770, opened 2025-10-30
Add metadata when using API to retrieve bundles


RomanFenioux:
**Is your feature request related to a problem? Please describe.**
I have developped a tool that allows user to interactively retrieve model-zoo bundles and run them on their medical scan within 3D slicer (an medical image processing and vizualisation application).
~~I retrieve the available model from the api : https://api.github.com/repos/Project-MONAI/model-zoo/releases/tags/hosting_storage_v1~~

EDIT :  I am now using the hugging_face api : https://huggingface.co/api/models?author=MONAI

I would like to present more information about the available models before downloading the bundles (which are heavy). Currently I only have access to the name and versions (extracted from the asset name) using the api linked above. In particular, I would like to retrieve the bundle description to help my users understand what each bundle can do.

I believe this would benefit anyone willing to provide access to the bundles from the Model Zoo through the api.

**Describe the solution you'd like**
My understanding is that this api uses the [model_info.json](https://github.com/Project-MONAI/model-zoo/blob/dev/models/model_info.json) file. I think adding the bundle description to this file would solve my problem, but maybe we could consider adding other relevant metadata (authors, name, version, task, copyright, dependencies...).

**Describe alternatives you've considered**
I could download the `metadata.json`  files for each model from the repo, but that would create many unnecessary request. It would be much cleaner to simply have access to this information in a single request, without downloading all those files.

Please let me now if there are other alternatives available to achieve my goal (retrieve the list of available models with their names, versions and description) that I might have failed to consider.

yiheng-wang-nv:
Hi @RomanFenioux , all latest versions are stored in huggingface instead of github release, it would be good if you can enhance new versions.

RomanFenioux:
Thank you, I had a look at the huggingface api and tested it.
The issue I raised about metadata still apply. Bundle metadatas are not exposed by the REST api (or through the huggingface_hub python package) so again the only solution is to download the `metadata.json` file.

PR #771, opened 2025-10-30
Add description metadata to model info

RomanFenioux:
### Description
This PR adds the description metadata to the model info file, as described in the issue #770  

The rationale is to have access to more information about the available model when using https://api.github.com/repos/Project-MONAI/model-zoo/releases/tags/hosting_storage_v1

Please let me know your thought, and if this is the right approach to solve this issue.

### Status
**Ready**

### DIFF
--- a/ci/update_model_info_deparate.py
+++ b/ci/update_model_info_deparate.py
@@ -84,10 +84,11 @@ def update_model_info(
     model_info = get_json_dict(model_info_path)
 
     if bundle_name_with_version not in model_info.keys():
-        model_info[bundle_name_with_version] = {"checksum": "", "source": ""}
+        model_info[bundle_name_with_version] = {"checksum": "", "source": "", "description": ""}
 
     model_info[bundle_name_with_version]["checksum"] = checksum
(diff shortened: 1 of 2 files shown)


ci/update_model_info_deparate.py
```diff
@@ -84,10 +84,11 @@ def update_model_info(
     model_info = get_json_dict(model_info_path)
 
     if bundle_name_with_version not in model_info.keys():
-        model_info[bundle_name_with_version] = {"checksum": "", "source": ""}
+        model_info[bundle_name_with_version] = {"checksum": "", "source": "", "description": ""}
```
yiheng-wang-nv, line 87: Hi @RomanFenioux , this script is an old version and is not used, can you revert the changes? Thanks

RomanFenioux: Thank you, i did not know that. Should the file be deleted to avoid confusion if the script is not used anymore ?

RomanFenioux: @yiheng-wang-nv when you have time, let me know if that would be ok with you
thanks !

